PR #1, opened 2025-02-26
feat(select) - build a simple select mimicking Remote styleguide 

gabrielseco:
## Overview

I buillt a Select component that we can use

This select component is composed by:

* form group which injects a label and a possible description text
* the select component itself

## Implementation decisions

I built the Select using the radix-ui/theme, I started with this one and give some css classes to the different parts.

There are some parts that are not accesible like the icon and that's why I am overriding classes

The Label is built with @radix-ui/react-label as the @radix-ui/themes doesn't provide a Label component

The CSS is built using custom properties, we have a primitives.css where can host shared variables between components and each component imports its own css

Our css classes are scoped with `rmt-SelectTrigger` and our css variables are scoped with `--rmt-select-trigger-padding`

Maybe all these decisions could be hosted in Architecture.MD file as this is the first approach let's keep in this PR and then decide

## How to use

I run a viteapp where I link the repo with `npm link` and in the App.tsx

```
import { Demo } from '@remoteoss/remote-flows/flows/Demo'

import './App.css'

function App() {

  return (
    <>
      <Demo />
    </>
  )
}

export default App

````

In the App.css

```
@import "node_modules/@remoteoss/remote-flows/dist/flows/Demo.css";
```

The only thing different that I have set up in the vitejs app is

```
import { defineConfig } from 'vite'
import react from '@vitejs/plugin-react-swc'
import path from 'path';

// https://vite.dev/config/
export default defineConfig({
  plugins: [react()],
  resolve: {
    alias: {
      react: path.resolve('./node_modules/react'),
    },
  },
})
```

I am aliasing react to avoid having an error about two react copies

## Demo

https://github.com/user-attachments/assets/eb90b43f-6ff0-4b2d-8649-764cb367eb7e

### DIFF
--- a/src/ui/Select.tsx
+++ b/src/ui/Select.tsx
@@ -1,0 +1,94 @@
+'use client';
+
+import * as React from 'react';
+import * as SelectPrimitive from '@radix-ui/react-select';
+
+const SelectRoot = SelectPrimitive.Root;
+
+const SelectValue = SelectPrimitive.Value;
+
+const SelectTrigger = React.forwardRef<
+  React.ComponentRef<typeof SelectPrimitive.Trigger>,
+  React.ComponentPropsWithoutRef<typeof SelectPrimitive.Trigger>
+>(({ className, children, ...props }, ref) => (
+  <SelectPrimitive.Trigger className={className} ref={ref} {...props}>
+    {children}
+  </SelectPrimitive.Trigger>
+));
+
+SelectTrigger.displayName = SelectPrimitive.Trigger.displayName;
+
+const SelectScrollUpButton = React.forwardRef<
+  React.ComponentRef<typeof SelectPrimitive.ScrollUpButton>,
+  React.ComponentPropsWithoutRef<typeof SelectPrimitive.ScrollUpButton>
+>(({ className, ...props }, ref) => (
+  <SelectPrimitive.ScrollUpButton
+    ref={ref}
+    {...props}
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/ui/Select.tsx [hunk 1]
```diff
@@ -0,0 +1,55 @@
...
+  const htmlFor = id || name;
+  return (
+    <FormGroup label={label} name={name} id={id} description={description}>
+      <SelectRadix.Root value={value} onValueChange={onChange}>
+        <SelectRadix.Trigger
+          id={htmlFor}
+          aria-describedby={`${htmlFor}-help-text`}
+          className="rmt-SelectTrigger"
+        />
+        <SelectRadix.Content className="rmt-SelectContent">
```
eng-almeida, line 41:
```suggestion
        <SelectRadix.Content className="rmt-SelectContent" position="popper">
```
We should do this update so that dropdown doesn't overlaps the dropdown. It's closer to our designs at Remote.

image

tsup.config.ts
```diff
@@ -6,7 +6,7 @@ export const tsup: Options = {
   target: 'esnext',
   clean: true,
   dts: true,
-  entry: ['src/index.tsx', 'src/flows'],
+  entry: ['src/index.tsx', 'src/flows/*.tsx'],
```
eng-almeida, line 9:
```suggestion
  entry: ['src/index.tsx', 'src/flows/*.ts*'],
```
I think it's better if we support both ts and tsx.

PR #5, opened 2025-03-03
feat(select): Themable select input

remotecom:
- Adds shadcn and Tailwind setup
- Adds shadcn `select` component
- Adds `ThemeProvider` to allow configuring a few CSS variables.

In the consumer app:

```
// App.css
@import "@remoteoss/remote-flows/index.css";
```

```jsx
  <ThemeProvider
    theme={{ colors: { primary: "red", input: "blue", accent: "green" } }}
  >
    <SelectField
      label="Select a country"
      options={countryOptions}
      onChange={(value) => console.log("Selected country:", value)}
    />
  </ThemeProvider>
```


src/types.ts
```diff
@@ -6,3 +9,28 @@ type AuthResponse = {
...
+        spacing?: string;
+        radius?: string;
+      }
+    | undefined;
+}>;
+
+export type ThemeColors = {
+  input: string;
+  primary: string;
+  accent: string;
```
remotecom, line 27: We can probably find better naming, so feel free to suggest anything 😄

gabrielseco:
how does one import the css from the package

I tried with App.css

@import "@remoteoss/remote-flows/dist/index.css";

but that didn't make the trick

gabrielseco:
I think for the first approach works, we probably need to dive deeper in the theming to be able to personalize all color for the select popover

image

PR #6, opened 2025-03-05
feat(select-field): Select as a form field

eng-almeida:
This PR wraps Select with shadcn FormField.

**How to test:**

Import `CostCalculator` in your consumer app, like so:

```jsx
import { CostCalculator } from "@remoteoss/remote-flows/flows/CostCalculator";

<ThemeProvider theme={{ colors: { dangerBackground: "black" } }}>
  <CostCalculator /> 
</ThemeProvider>
```

### DIFF
--- a/src/components/form/fields/SelectField.tsx
+++ b/src/components/form/fields/SelectField.tsx
@@ -17,7 +17,6 @@ type SelectFieldProps = {
   placeholder?: string;
   options: { value: string; label: string }[];
   defaultValue?: string;
-  onChange?: (value: string) => void;
   className?: string;
 };
 
@@ -26,7 +25,6 @@ export function SelectField({
   name,
   options,
   defaultValue,
-  onChange,
   className,
 }: SelectFieldProps) {
   const { control } = useFormContext();
@@ -40,13 +38,7 @@ export function SelectField({
         render={({ field, fieldState }) => (
           <FormItem>
             <FormControl>
-              <Select
-                value={field.value}
-                onValueChange={(value) => {
-                  field.onChange(value);
-                  onChange?.(value);
-                }}
-              >
+              <Select value={field.value} onValueChange={field.onChange}>
                 <SelectTrigger aria-invalid={Boolean(fieldState.error)}>
                   <span className={'text-foreground'}>
                     <SelectValue placeholder={label} />
--- a/src/components/form/fields/SelectField.tsx
+++ b/src/components/form/fields/SelectField.tsx
@@ -8,9 +8,12 @@ import {
   SelectTrigger,
   SelectValue,
 } from '@/src/components/ui/select';
+import { FormControl, FormField, FormItem, FormMessage } from '../../ui/form';
+import { useFormContext } from 'react-hook-form';
 
 type SelectFieldProps = {
   label: string;
+  name: string;
   placeholder?: string;
   options: { value: string; label: string }[];
   defaultValue?: string;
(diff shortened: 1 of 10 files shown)


src/types.ts
```diff
@@ -26,6 +26,8 @@ export type ThemeColors = Partial<{
   primaryForeground: string;
   accentBackground: string;
   accentForeground: string;
+  dangerBackground: string;
```
gabrielseco, line 29: I have in my demo

```
return (
    <ThemeProvider theme={{ colors: { dangerBackground: "black" } }}>
      <CostCalculator />
    </ThemeProvider>
  );
```

how should the app look?

image

even if I put red, I don't see any major change

src/components/ui/label.tsx
```diff
@@ -0,0 +1,24 @@
+"use client"
```
gabrielseco, line 1: **Question**: do we need the "use client"?

src/components/form/fields/SelectField.tsx
```diff
@@ -20,37 +23,50 @@ type SelectFieldProps = {
 
 export function SelectField({
   label,
+  name,
   options,
   defaultValue,
```
gabrielseco, line 28: do we need the onChange?

Asking because we'll use this with the form and the values will be stored in form

I'd prefer value but you could argue is an uncontrolled component if we rid of the onChange

src/components/form/fields/SelectField.tsx
```diff
@@ -20,37 +23,50 @@ type SelectFieldProps = {
...
   onChange,
   className,
 }: SelectFieldProps) {
-  const [value, setValue] = React.useState(defaultValue || '');
+  const { control } = useFormContext();
```
gabrielseco, line 32: nice didn't knew you could this, I was passing control as a prop in my branch

src/components/ui/form.tsx
```diff
@@ -0,0 +1,167 @@
...
+      id={formDescriptionId}
+      className={cn('text-muted-foreground text-sm', className)}
+      {...props}
+    />
+  );
+}
+
+function FormMessage({ className, ...props }: React.ComponentProps<'p'>) {
+  const { error, formMessageId } = useFormField();
+  console.log('###### ERROR', error);
```
gabrielseco, line 139: Nitpick: we can drop the log

src/components/ui/form.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const body = error ? String(error?.message ?? '') : props.children;
+
+  if (!body) {
+    return null;
+  }
+
+  return (
+    <p
+      data-slot="form-message"
```
gabrielseco, line 148: what are the purpose of the data-slot in this form

gabrielseco (commented):
I left some comments, testing the app now

gabrielseco:
Approved! there are not any major concerns.

Saying here "Approved" as the approve button doesn't appear

src/components/ui/label.tsx
eng-almeida: This code is generate by shadcn cli. We probably don't we need but it shouldn't cause any harm. The SDK should be able to work with SSR React components.

src/components/form/fields/SelectField.tsx
eng-almeida: We don't need it but I we might in the future. Will remove it for now.

src/components/ui/form.tsx
eng-almeida: This code is generate by shadcn cli. My interpretation of data-slot role is to provide an easy way to reference specific parts or slots of a component in a declarative manner.

remotecom (approved):

PR #9, opened 2025-03-06
feat(RadioGroup): Adds radio group field

eng-almeida:
Adds radio group component

### DIFF
--- a/src/components/form/fields/TextField.tsx
+++ b/src/components/form/fields/TextField.tsx
@@ -37,16 +37,18 @@ export function TextField({ name, description, label, type }: TextFieldProps) {
       <FormField
         control={control}
         name={name}
-        render={({ field }) => (
+        render={({ field, fieldState }) => (
           <FormItem className="TextField__Item">
             <FormLabel className="TextField__Label">{label}</FormLabel>
             <FormControl>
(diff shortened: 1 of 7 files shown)


src/components/form/fields/TextField.tsx
```diff
@@ -37,16 +37,18 @@ export function TextField({ name, description, label, type }: TextFieldProps) {
...
-            <FormDescription className="TextField__Description">
-              {description}
-            </FormDescription>
-            <FormMessage className="TextField__Error" />
+            {description && (
+              <FormDescription className="TextField__Description">
+                {description}
+              </FormDescription>
+            )}
+            {fieldState.error && <FormMessage className="TextField__Error" />}
```
remotecom, line 51: is there a reason of conditional rendering, I saw that FormMessage returns null when there are no errors.

remotecom (approved):
Just one comment

src/components/form/fields/TextField.tsx
eng-almeida: If we don't add this you'll get an empty <p> in the DOM, which isn't great. Plus it adds an extra vertical space because of line-height

PR #10, opened 2025-03-06
feat(fieldset): Add fieldset

eng-almeida:
Add fieldset field component

### DIFF
--- a/src/components/form/fields/FieldSetField.tsx
+++ b/src/components/form/fields/FieldSetField.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { RadioGroup } from '@/src/components/ui/radio-group';
 import { SupportedTypes } from './types';
 import { TextField } from './TextField';
-import { SelectField as FieldWithOptions } from './SelectField';
+import { SelectField } from './SelectField';
 
 type FieldBase = {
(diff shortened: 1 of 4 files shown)


src/components/form/fields/FieldSetField.tsx
```diff
@@ -0,0 +1,69 @@
...
+  fields: Field[];
+};
+
+export function FieldSetField({ legend, name, fields }: FieldSetProps) {
+  return (
+    <fieldset className="border-1 border-input p-4 rounded-xl">
+      <legend className="text-sm font-semibold px-2">{legend}</legend>
+      <div className="grid gap-4">
+        {fields.map((field) => {
+          if (field.type === 'select') {
```
gabrielseco, line 37: Wondered if we should do the same as in the remote-api-demo

https://github.com/remoteoss/remote-api-demo/blob/master/src/components/ui/form/Fields.jsx#L330

gabrielseco (approved):
only one comment

src/components/form/fields/FieldSetField.tsx
eng-almeida: Good idea! Code updated!

PR #12, opened 2025-03-10
feat(cost-calculator) - implement endpoints with react-query

gabrielseco:
## Overview

Implemented the cost calculator flow using react-query. I implemented the thing with ReactQueryProvider inserted in the RemoteFlows provider.

Haven't tested yet what happens with multiple providers in the page, but according to ChatGPT it should work

## How to test

in the App.tsx

```
import { useState } from "react";
import { RemoteFlows } from "@remoteoss/remote-flows";
import { CostCalculator } from "@remoteoss/remote-flows/flows/CostCalculator";
import "./App.css";

const ACCESS_TOKENS = {
  PARTNER: "<PARTNER_TOKEN>",
  NEW_COMPANY: "<COMPANY_TOKEN>",
};

function App() {
  const [estimations, setEstimations] = useState(null);
  return (
    <RemoteFlows
      auth={() =>
        Promise.resolve({
          expires_in: 3600,
          access_token: ACCESS_TOKENS.PARTNER,
        })
      }
    >
      <CostCalculator onSubmit={(response) => setEstimations(response)} />
      {estimations && <pre>{JSON.stringify(estimations, null, 2)}</pre>}
    </RemoteFlows>
  );
}

export default App;



````

In the App.css

```
@import "@remoteoss/remote-flows/index.css";
```

My vite.config.ts

```
import { defineConfig } from "vite";
import react from "@vitejs/plugin-react-swc";
import path from "path";

// https://vite.dev/config/
export default defineConfig({
  plugins: [react()],
  resolve: {
    alias: {
      react: path.resolve("./node_modules/react"),
    },
  },
  server: {
    proxy: {
      "/api": {
        target: "https://gateway.niceremote.com",
        changeOrigin: true,
        rewrite: (path) => path.replace(/^\/api/, ""),
      },
    },
  },
});

```

I am aliasing react to avoid having an error about two react copies

### DIFF
--- a/src/flows/CostCalculator.tsx
+++ b/src/flows/CostCalculator.tsx
@@ -165,7 +165,6 @@ export function CostCalculator({ onSubmit }: Props) {
 
   useEffect(() => {
     if (selectedCountry) {
-      console.log({ selectedCountry });
       if (
         selectedCountry?.childRegions.length === 0 &&
         selectedCountry.hasAdditionalFields
--- a/src/flows/CostCalculator.tsx
+++ b/src/flows/CostCalculator.tsx
@@ -129,7 +129,7 @@ export function CostCalculator({ onSubmit }: Props) {
     },
     mode: 'onBlur',
   });
-  const selectedCountry = form.watch('country');
+  const selectedCountryForm = form.watch('country');
   const selectedRegion = form.watch('region');
 
   const { data: currencies = [] } = useCompanyCurrencies();
@@ -137,11 +137,24 @@ export function CostCalculator({ onSubmit }: Props) {
   const { data: jsonSchemaForm } =
     useCalculatorLoadRegionFieldsSchemaForm(regionSlug);
 
+  const selectedCountry = useMemo(() => {
+    if (!selectedCountryForm) {
+      return null;
+    }
+
+    const country = countries?.find((c) => c.value === selectedCountryForm);
+
+    if (!country) {
+      return null;
+    }
+
+    return country;
+  }, [selectedCountryForm]);
+
   const regions = useMemo(() => {
     if (selectedCountry) {
-      const country = countries?.find((c) => c.value === selectedCountry);
       return (
-        country?.childRegions.map((region) => ({
+        selectedCountry?.childRegions.map((region) => ({
           value: region.slug,
           label: region.name,
         })) ?? []
@@ -152,11 +165,13 @@ export function CostCalculator({ onSubmit }: Props) {
 
   useEffect(() => {
     if (selectedCountry) {
-      const country = countries?.find((c) => c.value === selectedCountry);
-
-      if (country?.childRegions.length === 0 && country.hasAdditionalFields) {
+      console.log({ selectedCountry });
+      if (
+        selectedCountry?.childRegions.length === 0 &&
+        selectedCountry.hasAdditionalFields
+      ) {
         // test this with italy
-        setRegionSlug(country.regionSlug);
+        setRegionSlug(selectedCountry.regionSlug);
       }
     }
   }, [selectedCountry, countries]);
@@ -167,15 +182,10 @@ export function CostCalculator({ onSubmit }: Props) {
     }
   }, [selectedRegion]);
 
-  const findSlugInCountry = (countryCode: string) => {
-    const country = countries?.find((c) => c.value === countryCode);
-    return country?.regionSlug;
-  };
-
   const handleSubmit = (values: FormValues) => {
     const regionSlug = values.region
       ? values.region
-      : findSlugInCountry(values.country);
+      : selectedCountry?.regionSlug;
 
     const payload = {
       employer_currency_slug: values.currency as string,
--- a/src/flows/CostCalculator.tsx
+++ b/src/flows/CostCalculator.tsx
@@ -45,7 +45,6 @@ export function CostCalculator({ onSubmit }: Props) {
     validationSchema,
     handleJSONSchemaValidation,
   );
-  const [regionSlug, setRegionSlug] = useState<string | null>(null);
   const form = useForm<FormValues>({
     resolver: resolver,
     defaultValues: {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/flows/CostCalculator.tsx [hunk 1]
```diff
@@ -62,128 +145,63 @@ export function CostCalculator({ onSubmit }: Props) {
     },
     mode: 'onBlur',
   });
-  const selectedCountry = form.watch('country');
+  const selectedCountryForm = form.watch('country');
```
eng-almeida, line 148: Suggestion:
```suggestion
  const selectedCountry = form.watch('country');
```

src/flows/CostCalculator.tsx
```diff
@@ -62,128 +145,63 @@ export function CostCalculator({ onSubmit }: Props) {
...
-
-  const currenciesOptions = currencies.map(
-    (currency: { code: string; slug: string }) => ({
-      value: currency.slug,
-      label: currency.code,
-    }),
-  );
+  const { data: currencies = [] } = useCompanyCurrencies();
+  const { data: countries = [] } = useCalculatorCountries();
+  const { data: jsonSchemaForm } =
```
eng-almeida, line 153:
```suggestion
  const { data: jsonSchemaRegionFields } =
```

src/flows/CostCalculator.tsx
```diff
@@ -62,128 +145,63 @@ export function CostCalculator({ onSubmit }: Props) {
...
       }
     }
   }, [selectedCountry, countries]);
 
   useEffect(() => {
     if (selectedRegion) {
-      loadJsonForm(selectedRegion);
+      setRegionSlug(selectedRegion);
     }
   }, [selectedRegion]);
```
eng-almeida, line 199: Do we need this useEffect? I was able to create an estimation in the US which requires a region?

src/flows/CostCalculator.tsx
```diff
@@ -27,6 +26,91 @@ import { JSONSchemaFormFields } from '../components/form/JSONSchemaForm';
...
+          Authorization: ``,
+        },
+        body: payload,
+      });
+    },
+  });
+};
 
 const validationSchema = object({
   country: string().required('Country is required'),
```
eng-almeida, line 116: `salary` should be a number, so the schema should be updated to something along these lines:

```js
salary: number().typeError('Salary must be a number').required('Salary is required'),
```

src/flows/CostCalculator.tsx
```diff
@@ -27,6 +26,91 @@ import { JSONSchemaFormFields } from '../components/form/JSONSchemaForm';
 import { SelectField } from '@/src/components/form/fields/SelectField';
 import { TextField } from '@/src/components/form/fields/TextField';
 import { useValidationFormResolver } from '@/src/components/form/yupValidationResolver';
+import { useMutation, useQuery } from '@tanstack/react-query';
+
+const useCalculatorCountries = () => {
```
eng-almeida, line 31: Can we move the hooks to a separate file (maybe `hooks.ts`)? In this [MR](https://github.com/remoteoss/remote-flows/pull/13) I'm creating a CostCalculator folder, so we can have multiple files connected to cost calculator.

src/flows/CostCalculator.tsx
```diff
@@ -62,128 +145,63 @@ export function CostCalculator({ onSubmit }: Props) {
...
 
-      handleJSONSchemaValidation.current = jsonSchemaForm.handleValidation;
+    return country;
+  }, [selectedCountryForm]);
 
-      setJsonForm(jsonSchemaForm);
-    } catch (e) {
-      console.error(e);
-      setJsonForm(null);
+  const regions = useMemo(() => {
```
eng-almeida, line 171: question: does it pay off to memo regions? I'm not sure this is complex enough that justifies memozing the regions

eng-almeida (changes requested):

src/flows/CostCalculator.tsx
gabrielseco: Yes, when you select a region you need to be able to set the regionSlug to load the json schema form fields for the region

src/flows/CostCalculator.tsx
gabrielseco: I changed it

src/flows/CostCalculator.tsx
gabrielseco: done

gabrielseco (commented):
lol comments now go!!!!

src/flows/CostCalculator.tsx
gabrielseco: done

src/flows/CostCalculator.tsx
gabrielseco: done

src/flows/CostCalculator.tsx
```diff
@@ -27,6 +25,75 @@ import { JSONSchemaFormFields } from '../components/form/JSONSchemaForm';
 import { SelectField } from '@/src/components/form/fields/SelectField';
 import { TextField } from '@/src/components/form/fields/TextField';
 import { useValidationFormResolver } from '@/src/components/form/yupValidationResolver';
+import { useQuery } from '@tanstack/react-query';
+
+const useCalculatorCountries = () => {
```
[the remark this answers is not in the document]
gabrielseco: I think the hooks file, works at the moments

src/flows/CostCalculator.tsx
eng-almeida: Ah I see! But I think you can do this 
```js
const { data: jsonSchemaForm } = useCalculatorLoadRegionFieldsSchemaForm(
    form.watch('region'),
  );
```
instead of having a dedicated useEffect for this.

gabrielseco: Oh that will work or I hope so

gabrielseco: yes it works much better now

eng-almeida (approved):

PR #14, opened 2025-03-11
fix(radio): fix indicator color

eng-almeida:
Add radio color to be used by radio indicator so that by default color of the "circle" isn't white (primary color)

### DIFF
--- a/src/components/form/fields/RadioGroupField.tsx
+++ b/src/components/form/fields/RadioGroupField.tsx
@@ -1,5 +1,3 @@
-import * as React from 'react';
-import { useFormContext } from 'react-hook-form';
 import {
   FormControl,
   FormDescription,
@@ -9,6 +7,8 @@ import {
   FormMessage,
 } from '@/src/components/ui/form';
 import { RadioGroup, RadioGroupItem } from '@/src/components/ui/radio-group';
+import * as React from 'react';
+import { useFormContext } from 'react-hook-form';
 
 type RadioGroupFieldProps = {
   name: string;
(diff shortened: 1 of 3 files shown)


src/components/form/fields/RadioGroupField.tsx
```diff
@@ -25,6 +25,7 @@ export function RadioGroupField({
   label,
   options,
 }: RadioGroupFieldProps) {
+  console.log(options);
   const { control } = useFormContext();
```
gabrielseco, line 29: remove log, we probably need an eslint to avoid this kind of problems in the review.

src/components/ui/radio-group.tsx
```diff
@@ -34,7 +34,7 @@ function RadioGroupItem({
         data-slot="radio-group-indicator"
         className="relative flex items-center justify-center"
       >
-        <CircleIcon className="fill-primary absolute top-1/2 left-1/2 size-2 -translate-x-1/2 -translate-y-1/2" />
+        <CircleIcon className="fill-radio absolute top-1/2 left-1/2 size-2 -translate-x-1/2 -translate-y-1/2" />
```
gabrielseco, line 37: Question: does the fill-radio magically added by tailwind when adding the --radio property?

gabrielseco (commented):
Add some comments after code review, checking in browser

src/components/ui/radio-group.tsx
eng-almeida: Yes! Tailwind automatically generates CSS classes that include the new color you are adding. To make it work, you just need to create a CSS variable that starts with `--color`. For example, `--color-hello`, will generate class for text color like `text-hello` or background color `bg-hello`.

PR #20, opened 2025-03-12
feat(demo) - add example folder with vite and express

gabrielseco:
## Overview

Implemented a demo app in the example folder, it's built using vite && express together

## Caveats

* I had difficulties running express with typescript, it wasn't playing nice with tsconfig file, I put a javascript file for now.
* in the vitejs config, react is aliased as I was having problems due to linking with file:../ and react wasn't running
* Now that the token is fetched, we probably need a loading mechanism to let know the user that we're fetching data from the browser
* I see the token is fetched twice, maybe something related to the auth function and strict mode, maybe?

## How to use

```
 npm run dev
```

after that 

```
CLIENT_ID=<CLIENT_ID>
CLIENT_SECRET=<CLIENT_ID>
```

```

cd example
npm run dev
```

Open localhost:3000 and the app should work

https://github.com/user-attachments/assets/f541df01-541d-4f27-bc2e-4982b23dacbf


eng-almeida:
@gabrielremote I understand why we're fetching /token multiple times. We had 2 different issues:
- Parallel requests, like we have in cost calculator `/countries` and `/currencies` where both fetching the token endpoint. Using tanstack query, fixes the duplicated requests.
- I haven't calculated  correctly `expiresAt` when I implemented this 🤦 This should prevent new token fetch until the token expires. For example, you can load the cost calculator and if you select `Italy`, which triggers the regions endpoint, the token endpoint won't be fetched again.

I'm pasting here the fix:

```jsx
import { Client, createClient } from '@hey-api/client-fetch';
import {
  QueryClient,
  QueryClientProvider,
  useQuery,
} from '@tanstack/react-query';
import type { PropsWithChildren } from 'react';
import React, { createContext, useContext, useRef } from 'react';

import { client } from './client/client.gen';
import { RemoteFlowsSDKProps } from './types/remoteFlows';

const RemoteFlowContext = createContext<{ client: Client | null }>({
  client: null,
});

const queryClient = new QueryClient();

export const useClient = () => useContext(RemoteFlowContext);

function RemoteFlowContextWrapper({ children, auth }: any) {
  const session = useRef<{ accessToken: string; expiresAt: number } | null>(
    null,
  );
  const { refetch } = useQuery<{ access_token: string; expires_in: string }>({
    queryKey: ['auth'],
    queryFn: auth,
    enabled: false,
  });
  const remoteApiClient = useRef(
    createClient({
      ...client.getConfig(),
      baseUrl: process.env.REMOTE_GATEWAY_URL,
      auth: async () => {
        function hasTokenExpired(expiresAt: number | undefined) {
          return !expiresAt || Date.now() + 60000 > expiresAt;
        }

        if (!session.current || hasTokenExpired(session.current.expiresAt)) {
          const { data } = await refetch();
          if (data) {
            session.current = {
              accessToken: data.access_token,
              expiresAt: Date.now() + parseInt(data.expires_in, 10) * 1000,
            };
          }
        }
        return session.current?.accessToken;
      },
    }),
  );
  return (
    <RemoteFlowContext.Provider value={{ client: remoteApiClient.current }}>
      {children}
    </RemoteFlowContext.Provider>
  );
}

export function RemoteFlows({
  auth,
  children,
}: PropsWithChildren<RemoteFlowsSDKProps>) {
  return (
    <QueryClientProvider client={queryClient}>
      <RemoteFlowContextWrapper auth={auth}>
        {children}
      </RemoteFlowContextWrapper>
    </QueryClientProvider>
  );
}
```

src/types/remoteFlows.ts
```diff
@@ -1,7 +1,7 @@
 // auth types
 type AuthResponse = {
   access_token: string;
-  expires_in: number;
+  expires_in: string;
```
eng-almeida, line 4: I know I was the one adding this but can you update these keys to camelcase? 🙏

eng-almeida (approved):

src/types/remoteFlows.ts
gabrielseco: ok, I didn't know if it was intentional or not that's why I didn't change anything

gabrielseco: I applied the change, actually I revert the type to number as we receive a number

eng-almeida (approved):

PR #23, opened 2025-03-13
feat(estimation) - export pdf

gabrielseco:
## Overview

Implemented the pdf export, a button appears after you submit the form data.

I offer the users a hook, where they can send the payload the submitted and with that a pdf will be downloaded.

## Caveats

* I believe that I have to tweak the Results component and transformFromCents

## How to use

```
 npm run dev
```

after that 

```
CLIENT_ID=<CLIENT_ID>
CLIENT_SECRET=<CLIENT_ID>
```

```

cd example
npm run dev
```

Open localhost:3000 and the app should work

https://github.com/user-attachments/assets/f97b0296-f0ec-494f-84fe-2f27d974a0ef


example/server.js
```diff
@@ -40,7 +40,7 @@ const getToken = async (req, res) => {
 
 const startServer = async () => {
   const app = express();
-  const port = 3000;
+  const port = 3001;
```
gabrielseco, line 43: change the port to avoid conflicts with dragon

src/flows/CostCalculator/exportAsPdf.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  const { mutate } = exportAsPDFMutation;
+
+  return useCallback(
+    (data: CostCalculatorEstimateParams) => {
+      return new Promise<string>((resolve, reject) => {
+        mutate(data, {
+          onSuccess: (data) => {
+            if (data.data?.data.content !== undefined) {
+              resolve(data.data?.data.content as $TSFixMe);
```
gabrielseco, line 17: I don't know if the type that returns the mutation is the correct one

eng-almeida:
@gabrielseco `package-lock.json` has changes even though there's no changes in `package.json`. Can you please revert it? 🙏

src/flows/CostCalculator/hooks.ts
```diff
@@ -249,3 +267,27 @@ export const useCostCalculator = (): BaseHookReturn<
...
+          },
+          onError: (error) => {
+            reject(error);
+          },
+        });
+      });
+    },
+    [mutate],
+  );
+};
```
eng-almeida, line 293: IMO this is consumer responsibility so I don't think we need this hook. Consumer apps can do whatever they want/need with `useCostCalculatorEstimationPdf`

src/flows/CostCalculator/hooks.ts
```diff
@@ -99,6 +101,22 @@ const useCostCalculatorEstimation = () => {
   });
 };
 
+export const useCostCalculatorEstimationPdf = () => {
```
eng-almeida, line 104: Can you add a simple JS docs like we have for the other hooks? We'll have to improve this docs some other time 👍

eng-almeida (changes requested):

src/flows/CostCalculator/hooks.ts
gabrielseco: ok, let's simplify

eng-almeida (approved):

PR #24, opened 2025-03-14
feat(baseUrl) - change the API baseUrl based on the environment

gabrielseco:
I was having doubts if we should allow whatever string or just let people choose between staging / production.

i choose the latter

### DIFF
--- a/src/RemoteFlowsProvider.tsx
+++ b/src/RemoteFlowsProvider.tsx
@@ -21,11 +21,18 @@ export const useClient = () => useContext(RemoteFlowContext);
 type RemoteFlowContextWrapperProps = {
   auth: RemoteFlowsSDKProps['auth'];
   children: React.ReactNode;
+  environment?: RemoteFlowsSDKProps['environment'];
+};
+
+const ENVIROMENTS = {
+  staging: 'https://gateway.niceremote.com/',
+  production: 'https://gateway.remote.com/',
 };
 
 function RemoteFlowContextWrapper({
   children,
   auth,
+  environment = 'staging',
 }: RemoteFlowContextWrapperProps) {
   const session = useRef<{ accessToken: string; expiresAt: number } | null>(
     null,
@@ -35,10 +42,13 @@ function RemoteFlowContextWrapper({
     queryFn: auth,
     enabled: false,
   });
+
+  const baseUrl = ENVIROMENTS[environment as keyof typeof ENVIROMENTS];
+
   const remoteApiClient = useRef(
     createClient({
       ...client.getConfig(),
-      baseUrl: process.env.REMOTE_GATEWAY_URL,
+      baseUrl,
       auth: async () => {
         function hasTokenExpired(expiresAt: number | undefined) {
           return !expiresAt || Date.now() + 60000 > expiresAt;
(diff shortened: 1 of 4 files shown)


tsup.config.ts
```diff
@@ -1,7 +1,7 @@
 import { Options } from 'tsup';
 
-const env = process.env.NODE_ENV;
-
+/* const env = process.env.NODE_ENV;
```
gabrielseco, line 3: We can delete this right?

And the line 15

src/RemoteFlowsProvider.tsx
```diff
@@ -21,11 +21,18 @@ export const useClient = () => useContext(RemoteFlowContext);
 type RemoteFlowContextWrapperProps = {
   auth: RemoteFlowsSDKProps['auth'];
   children: React.ReactNode;
+  environment?: RemoteFlowsSDKProps['environment'];
+};
+
+const ENVIROMENTS = {
```
gabrielseco, line 27: let me know if you want to move this to a constants file

src/RemoteFlowsProvider.tsx
```diff
@@ -35,10 +42,13 @@ function RemoteFlowContextWrapper({
     queryFn: auth,
     enabled: false,
   });
+
+  const baseUrl = ENVIROMENTS[environment as keyof typeof ENVIROMENTS];
+
   const remoteApiClient = useRef(
     createClient({
       ...client.getConfig(),
-      baseUrl: process.env.REMOTE_GATEWAY_URL,
```
eng-almeida, line 41: I think it's preferable to not remove this. As it stands this would mean that `environment` is mandatory prop which I don't think it should be. IMO `environment` should be optional and not have a default value. 
This would allow us to have this: `baseUrl: environment || process.env.REMOTE_GATEWAY_URL`

Let me know your thoughts

gabrielseco: and the environment prop is whatever string or it can be the ones that I already defined?

eng-almeida: that's ok, it should be either `staging` or `production`

eng-almeida:
@gabrielremote what do you think if instead of using `environment` prop we use `isTestingMode`? Which means that when it's set to `true` we use staging environment? The reason I'm asking it's because it feels to me that this name is self explanatory.

gabrielseco:
I was setting environment in case we have sandbox in the future?

Does the sandbox use the same baseURL as staging?

cc: @eng-almeida 
> @gabrielremote what do you think if instead of using `environment` prop we use `isTestingMode`? Which means that when it's set to `true` we use staging environment? The reason I'm asking it's because it feels to me that this name is self explanatory.

eng-almeida:
@gabrielremote can partners use sandbox? 🤔

src/RemoteFlowsProvider.tsx
eng-almeida: Yeah, what do you think of creating a `src/environment.ts`? `tsup.config.ts` should also import it.

gabrielseco: it's done

eng-almeida (approved):

PR #25, opened 2025-03-14
feat(disclaimer) - add disclaimer to the cost calculator

gabrielseco:
## Overview

Implemented a disclaimer component that is rendered inside the cost calculator, we also offer a hook, that way partners can rely on the data and be more opinionated about the styles.

## Implementation details
In the example folder, I added some css on how one can style the drawer.

Add new dependencies from shadcn to load drawers.

## How to use

```
 npm run dev
```

after that 

```
CLIENT_ID=<CLIENT_ID>
CLIENT_SECRET=<CLIENT_ID>
```

```

cd example
npm run dev
```

Open localhost:3001 and the app should work

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
   "dependencies": {
     "@hey-api/client-fetch": "^0.8.1",
     "@hookform/resolvers": "^4.1.3",
+    "@radix-ui/react-dialog": "^1.1.6",
     "@radix-ui/react-label": "^2.1.2",
     "@radix-ui/react-radio-group": "^1.2.3",
     "@radix-ui/react-select": "^2.1.6",
@@ -67,6 +68,7 @@
     "tailwindcss": "^4.0.9",
     "tailwindcss-animate": "^1.0.7",
     "type-fest": "^4.35.0",
+    "vaul": "^1.1.2",
     "yup": "^0.32.0",
     "zod": "^3.24.2"
   },
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "@radix-ui/react-dialog": "^1.1.6",
     "@radix-ui/react-label": "^2.1.2",
     "@radix-ui/react-radio-group": "^1.2.3",
+    "@radix-ui/react-scroll-area": "^1.2.3",
     "@radix-ui/react-select": "^2.1.6",
     "@radix-ui/react-separator": "^1.1.2",
     "@radix-ui/react-slot": "^1.1.2",
--- a/package.json
+++ b/package.json
@@ -70,8 +70,7 @@
     "tailwindcss-animate": "^1.0.7",
     "type-fest": "^4.35.0",
     "vaul": "^1.1.2",
-    "yup": "^0.32.0",
-    "zod": "^3.24.2"
+    "yup": "^0.32.0"
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.17.3",
--- a/src/components/ui/button.tsx
+++ b/src/components/ui/button.tsx
@@ -18,7 +18,7 @@ const buttonVariants = cva(
         secondary:
           'bg-secondary text-secondary-foreground shadow-xs hover:bg-secondary/80',
         ghost: 'hover:bg-accent hover:text-accent-foreground',
-        link: 'text-primary-foreground underline-offset-4 hover:underline button-link',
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -67,6 +69,7 @@
     "tailwindcss": "^4.0.9",
     "tailwindcss-animate": "^1.0.7",
     "type-fest": "^4.35.0",
+    "vaul": "^1.1.2",
```
gabrielseco, line 72: shadcn uses vaul https://ui.shadcn.com/docs/components/drawer

src/components/ui/button.tsx
```diff
@@ -18,13 +18,14 @@ const buttonVariants = cva(
         secondary:
           'bg-secondary text-secondary-foreground shadow-xs hover:bg-secondary/80',
         ghost: 'hover:bg-accent hover:text-accent-foreground',
-        link: 'text-primary underline-offset-4 hover:underline',
+        link: 'text-primary-foreground underline-offset-4 hover:underline button-link',
```
gabrielseco, line 21: I changed link variant as text-primary is white.

I wonder if we should add classes like button-link, button-ghost

that way partners can override buttons easily

src/components/ui/button.tsx
```diff
@@ -18,13 +18,14 @@ const buttonVariants = cva(
...
       },
       size: {
         default: 'h-9 px-4 py-7 has-[>svg]:px-3',
         sm: 'h-8 rounded-md gap-1.5 px-3 has-[>svg]:px-2.5',
         lg: 'h-10 rounded-md px-6 has-[>svg]:px-4',
         icon: 'size-9',
+        link: 'px-0',
```
gabrielseco, line 28: the button has always padding which you don't want in links

src/flows/CostCalculator/Disclaimer/Disclaimer.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { ScrollArea } from '@/src/components/ui/scroll-area';
+import { X } from 'lucide-react';
+import { Button } from '@/src/components/ui/button';
+import { useCostCalculatorDisclaimer } from '@/src/flows/CostCalculator/hooks';
+
+type DisclaimerProps = {
+  label?: string;
+};
+
+export const Disclaimer = ({ label = 'Disclaimer' }: DisclaimerProps) => {
```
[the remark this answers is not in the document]
gabrielseco: what if we move the Disclaimer to the example folder and we leave the code there to let partners know how to implement.

What do you think?

I don't mind deleting.

We can ask Umbreen about it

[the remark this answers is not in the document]
eng-almeida: Good idea, let's move it to the example folder 👍 But let's ask Umbreen as well, as I might be wrong here 🙂

src/components/ui/button.tsx
eng-almeida: > I changed link variant as text-primary is white.

We shouldn't do this because that would break [Shadcn naming convention](https://ui.shadcn.com/docs/theming) 
image

If we need to use change the text color for the link buttons, I think it's preferable to create a new color in global.css and follow the convention

```css

--link-button: var(--primaryForeground, #ffffff);

@theme inline {
  --color-link-button-primary: var(--link-button)
}

```

```js
link: 'text-link-button-primary underline-offset-4 hover:underline button-link'
```

Anyway, if we're moving the disclaimer implementation to the example folder, we won't need this for now.

remotecom: I created the link-button using the foreground as the foreground is for text color.

I am wondering what's the difference between 

text-primary-foreground and the --link-button isn't the same thing?

it seems to me that we're using the same colors

eng-almeida: Yes, they are the same colours. I've made some confusion with the color you're changing, sorry! If I understand this correctly you're replacing the background color with the text color. Is this what you really want? Feel free to ignore this comment, if that's intended.
My suggestion it's for the cases when in a component you want to use the background color as the text color or the other way around.

gabrielseco: yes because if text and background have same colors, you won't see the text, we can leave this new class for now

eng-almeida (approved):

PR #26, opened 2025-03-17
docs(cost-calculator) - throw errors correctly

gabrielseco:
I was adding the test for the onError and I notice that the code wasn't executing correctly.

It seems like @hey-client/api doesn't throw and instead offers a response with {error, request, response}

What I did is whenever it returns a status code error, throw the error to make the CostCalculator component to catch it and emit the error with onError callback

### DIFF
--- a/src/flows/CostCalculator/hooks.ts
+++ b/src/flows/CostCalculator/hooks.ts
@@ -11,7 +11,7 @@ import {
 import type { BaseHookReturn, Field } from '@/src/flows/CostCalculator/types';
 import { useClient } from '@/src/RemoteFlowsProvider';
 import { Client } from '@hey-api/client-fetch';
-import { createHeadlessForm } from '@remoteoss/json-schema-form';
+import { $TSFixMe, createHeadlessForm } from '@remoteoss/json-schema-form';
 import { useMutation, useQuery } from '@tanstack/react-query';
 import { useState } from 'react';
 import { AnyObjectSchema, object, string } from 'yup';
@@ -185,6 +185,13 @@ export const useCostCalculator = (): BaseHookReturn<
     values: CostCalculatorEstimateParams,
   ): Promise<CostCalculatorEstimateResponse | null> {
     const response = await costCalculatorEstimationMutation.mutateAsync(values);
+    if (
+      response.error &&
+      response.response.status >= 400 &&
+      response.response.status < 600
+    ) {
+      throw new Error((response.error as $TSFixMe).message);
+    }
     if (response.data) {
       return response.data;
     }
--- a/src/flows/CostCalculator/tests/CostCalculator.test.tsx
+++ b/src/flows/CostCalculator/tests/CostCalculator.test.tsx
@@ -91,21 +91,20 @@ describe('CostCalculator', () => {
     });
   });
 
-  // test('calls onSuccess when estimation succeeds', async () => {
-  //   render(<CostCalculator {...defaultProps} />, { wrapper });
+  test('calls onError when estimation fails', async () => {
+    server.use(
+      http.post('*/v1/cost-calculator/estimation', () => {
+        return new HttpResponse('Not found', { status: 404 });
+      }),
+    );
+    render(<CostCalculator {...defaultProps} />, { wrapper });
 
-  //   fireEvent.click(screen.getByRole('button', { name: /save/i }));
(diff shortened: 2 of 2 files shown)


src/flows/CostCalculator/hooks.ts
```diff
@@ -158,20 +158,30 @@ export const defaultEstimationOptions = {
   includeCostBreakdowns: false,
 };
 
+type UseCostCalculatorParams = {
+  defaultRegion?: string;
```
eng-almeida, line 162: Why not passing all the default values to the hook?

src/flows/CostCalculator/hooks.ts
```diff
@@ -158,20 +158,30 @@ export const defaultEstimationOptions = {
...
 /**
  * Hook to use the cost calculator.
  */
 export const useCostCalculator = (
-  estimationOptions: CostCalculatorEstimationOptions = defaultEstimationOptions,
+  { defaultRegion, estimationOptions }: UseCostCalculatorParams = {
+    estimationOptions: defaultEstimationOptions,
+  },
 ) => {
   const [selectedRegion, setSelectedRegion] = useState<string>();
```
eng-almeida, line 174: `useState` can take the region as the default value.

src/flows/CostCalculator/hooks.ts
```diff
@@ -158,20 +158,30 @@ export const defaultEstimationOptions = {
...
   const [selectedCountry, setSelectedCountry] =
     useState<CostCalculatorCountry>();
   const { data: countries } = useCostCalculatorCountries();
   const { data: currencies } = useCompanyCurrencies();
-  const { data: jsonSchemaRegionFields } = useRegionFields(
-    selectedRegion || selectedCountry?.value,
-  );
+
+  const jsonSchemaRegionSlug =
+    selectedRegion || selectedCountry?.value || defaultRegion;
```
eng-almeida, line 181: Assuming you'll follow my suggestion above, we can rollback this change

```suggestion
  const jsonSchemaRegionSlug =
    selectedRegion || selectedCountry?.value;
```

src/flows/CostCalculator/hooks.ts
```diff
@@ -195,6 +205,14 @@ export const useCostCalculator = (
         buildPayload(values, estimationOptions),
       );
 
+      if (
+        response.error &&
+        response.response.status >= 400 &&
+        response.response.status < 600
+      ) {
+        throw new Error((response.error as $TSFixMe).message);
+      }
```
eng-almeida, line 214: Why don't we keep @hey-client/api approach and keep it simpler? If the API throws an error, onError should be called now.

```js
   const response = await costCalculatorEstimationMutation.mutateAsync(
      buildPayload(values, estimationOptions),
    );

    if (response.data) {
      return {
        data: response.data,
        error: null,
      };
    }

    return {
      data: null,
      error: response.error,
    };
```

src/flows/CostCalculator/hooks.ts
gabrielseco: you only need the region to load dynamically the region fields.

What I was trying to solve is when you preselect Japan the regions are loaded automatically.

Salary and currency don't have this problem

src/flows/CostCalculator/hooks.ts
gabrielseco: done

src/flows/CostCalculator/hooks.ts
gabrielseco: done

src/flows/CostCalculator/hooks.ts
gabrielseco: done

src/flows/CostCalculator/hooks.ts
```diff
@@ -158,20 +158,31 @@ export const defaultEstimationOptions = {
...
+  );
   const [selectedCountry, setSelectedCountry] =
     useState<CostCalculatorCountry>();
   const { data: countries } = useCostCalculatorCountries();
   const { data: currencies } = useCompanyCurrencies();
-  const { data: jsonSchemaRegionFields } = useRegionFields(
-    selectedRegion || selectedCountry?.value,
-  );
+
+  const jsonSchemaRegionSlug = selectedRegion || selectedCountry?.value;
```
eng-almeida, line 182:
```suggestion
  const region = selectedRegion || selectedCountry?.value;
```

PR #27, opened 2025-03-17
fix(cost-calculator) - load region fields if a country has them

gabrielseco:
I noticed when you set a country with defaultValues that had region fields like Japan, the JSON schema wasn't getting fetched.

I fixed it and add some tests as I discovered this problem when I was doing some.

image

### DIFF
--- a/src/flows/CostCalculator/CostCalculator.tsx
+++ b/src/flows/CostCalculator/CostCalculator.tsx
@@ -95,7 +95,9 @@ export function CostCalculator({
     onSubmit: submitCostCalculator,
     fields,
     validationSchema,
-  } = useCostCalculator();
+  } = useCostCalculator({
+    countryRegionSlug: defaultValues.countryRegionSlug,
+  });
 
   const resolver = useValidationFormResolver(validationSchema);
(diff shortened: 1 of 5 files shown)


src/flows/CostCalculator/CostCalculator.tsx
```diff
@@ -95,7 +95,9 @@ export function CostCalculator({
     onSubmit: submitCostCalculator,
     fields,
     validationSchema,
-  } = useCostCalculator();
+  } = useCostCalculator({
+    countryRegionSlug: defaultValues.countryRegionSlug,
```
eng-almeida, line 99: I think it's easier to read this if we update `useCostCalculator` param from `countryRegionSlug` to `defaultRegion`

eng-almeida:
Nice catch! I left a minor suggestion 🙏

src/flows/CostCalculator/CostCalculator.tsx
gabrielseco: done!

eng-almeida (approved):

PR #32, opened 2025-03-19
chore(useCostCalculator): Move form validation to onSubmit

eng-almeida:
`onSubmit` function in `useCostCalculator` should validate form fields before posting the estimation to Remote API. 
If partners use `useCostCalculator` directly they might not have a validation in place, so this change ensures that fields validation is always executed before sending the request.

Beside including the validation, this PR also changes the `onSubmit` parameters. The hook should be responsible for formatting the request payload, not the component.

### DIFF
--- a/src/flows/CostCalculator/CostCalculator.tsx
+++ b/src/flows/CostCalculator/CostCalculator.tsx
@@ -1,48 +1,27 @@
 import React from 'react';
 import { useForm } from 'react-hook-form';
 
-import type {
-  CostCalculatorEstimateParams,
-  CostCalculatorEstimateResponse,
-  EmploymentTermType,
-} from '@/src/client';
+import type { CostCalculatorEstimateResponse } from '@/src/client';
 import { Form } from '@/src/components/ui/form';
 
 import { JSONSchemaFormFields } from '@/src/components/form/JSONSchemaForm';
 import { useValidationFormResolver } from '@/src/components/form/yupValidationResolver';
 import { Button } from '@/src/components/ui/button';
 import { Disclaimer } from '@/src/flows/CostCalculator/Disclaimer';
-import { useCostCalculator } from '@/src/flows/CostCalculator/hooks';
-import { convertToCents } from '@/src/lib/utils';
-
-type FormValues = {
-  currency: string;
-  country: string;
-  salary: string;
-} & Partial<{
-  region: string;
-  age: number;
-  contract_duration_type: EmploymentTermType;
-}>;
+import {
+  defaultEstimationOptions,
+  useCostCalculator,
+} from '@/src/flows/CostCalculator/hooks';
+import {
+  CostCalculatorEstimateFormValues,
+  CostCalculatorEstimationOptions,
+} from './types';
 
 type CostCalculatorProps = Partial<{
   /**
    * Estimation params allows you to customize the parameters sent to the /cost-calculator/estimation endpoint.
    */
-  estimationParams: Partial<{
-    /**
-     * Title of the estimation. Default is 'Estimation'.
-     */
-    title: string;
-    /**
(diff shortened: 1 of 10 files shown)


src/flows/CostCalculator/CostCalculator.tsx
```diff
@@ -1,48 +1,27 @@
...
-    /**
-     * Include benefits in the estimation. Default is false.
-     */
-    includeBenefits: boolean;
-    /**
-     * Include cost breakdowns in the estimation. Default is false.
-     */
-    includeCostBreakdowns: boolean;
-  }>;
+  estimationOptions?: CostCalculatorEstimationOptions;
```
eng-almeida, line 24: IMO `estimationOptions` is slightly better name than `estimationParams`. lmk your thoughts

src/flows/CostCalculator/CostCalculator.tsx
```diff
@@ -116,35 +91,15 @@ export function CostCalculator({
...
-          regional_to_employer_exchange_rate: '1',
-          age: values.age ?? undefined,
-        },
-      ],
-    };
+  const handleSubmit = async (values: CostCalculatorEstimateFormValues) => {
+    await onSubmit?.(values);
 
-    await onSubmit?.(payload);
+    const estimation = await submitCostCalculator(values);
```
gabrielseco, line 97: I was doing more or less the same as I had to offer to the hook form the ability to build a payload

src/index.tsx
```diff
@@ -5,9 +5,14 @@ export {
...
   CostCalculatorResults,
 } from '@/src/flows/CostCalculator';
+
 export {
-  useCostCalculatorEstimationPdf,
   useCostCalculatorDisclaimer,
+  useCostCalculatorEstimationPdf,
 } from '@/src/flows/CostCalculator/hooks';
+
+export { transformYupErrorsIntoObject } from '@/src/lib/utils';
```
gabrielseco, line 14: can you show how do you use this from a demo app or the example folder

gabrielseco (commented):
Leaving some comments

gabrielseco (approved):

src/flows/CostCalculator/CostCalculator.tsx
```diff
@@ -116,35 +91,15 @@ export function CostCalculator({
...
-          ),
-          employment_term: values.contract_duration_type ?? 'fixed',
-          title: estimationParams.title,
-          regional_to_employer_exchange_rate: '1',
-          age: values.age ?? undefined,
-        },
-      ],
-    };
+  const handleSubmit = async (values: CostCalculatorEstimateFormValues) => {
+    await onSubmit?.(values);
```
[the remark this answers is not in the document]
eng-almeida: Yes, it's easier to digest. I'm assuming consumer apps are more interested in the form payload than the formatted payload that Remote API consumes.

src/index.tsx
eng-almeida: transformYupErrorsIntoObject is only valuable when consumers app use `useCostCalculator` hook instead of `CostCalculator` component. In CostCalculator we use react-hook-form and the built in validations, so when onSubmit runs the payload is always valid.

```js
import { transformYupErrorsIntoObject } from '@remoteoss/remote-flows'

function MyComponent() {
  const {
    onSubmit,
    fields,
    validationSchema,
  } = useCostCalculator(estimationOptions);

  const handleSubmit = async (values: CostCalculatorEstimationFormValues) => {
    const estimation = await onSubmit(values);

    if (estimation.error) {
      console.log(transformYupErrorsIntoObject(estimation.error)) // with this object consumers can manually set validations if they want to
    } 
  };
  
  return <form onSubmit={handleSubmit}></form>
}
```

PR #40, opened 2025-03-21
feat(CostCalculator): support isLoading and isSubmitting

eng-almeida:
This PR adds `isLoading` and `isSubmitting` to `useCostCalculator` hook.
- `isLoading` state derives from loading countries, currencies and region field
- `isSubmitting` state derives from the mutation isPending state

It also adds `useCostCalculator` documentation and some minor CSS improvements to the cost calculator example

### DIFF
--- a/src/flows/CostCalculator/CostCalculator.tsx
+++ b/src/flows/CostCalculator/CostCalculator.tsx
@@ -10,6 +10,7 @@ import { Button } from '@/src/components/ui/button';
 import { Disclaimer } from '@/src/flows/CostCalculator/Disclaimer';
 import {
   defaultEstimationOptions,
+  EstimationError,
   useCostCalculator,
 } from '@/src/flows/CostCalculator/hooks';
 
@@ -69,7 +70,7 @@ type CostCalculatorProps = Partial<{
    * Callback function to handle the error when the estimation fails.
    * @param error - The error object.
    */
-  onError: (error: Error) => void;
+  onError: (error: EstimationError) => void;
 }>;
 
 export function CostCalculator({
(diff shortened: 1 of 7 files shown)


src/flows/CostCalculator/CostCalculator.tsx
```diff
@@ -128,8 +130,9 @@ export function CostCalculator({
           <Button
             type="submit"
             className="w-full bg-gray-900 hover:bg-gray-800 text-white"
+            disabled={isSubmitting}
           >
-            Save
+            Get estimate
```
eng-almeida, line 135: Sounds like a better name. We should probably think about making this configurable

src/flows/CostCalculator/hooks.ts
```diff
@@ -201,28 +206,34 @@ export const useCostCalculator = (
       };
     }
 
-    try {
-      const response = await costCalculatorEstimationMutation.mutateAsync(
+    return new Promise((resolve, reject) => {
+      costCalculatorEstimationMutation.mutate(
```
eng-almeida, line 210: Changed from `mutateAsync` to `mutate` so we can levarage `costCalculatorEstimationMutation.isPending` instead of manually handling custom state

src/flows/CostCalculator/CostCalculator.tsx
gabrielseco: yes, probably with the options prop

src/flows/CostCalculator/hooks.ts
gabrielseco: does the isPending don't work with mutateAsync?

gabrielseco (approved):
left two comments

src/flows/CostCalculator/hooks.ts
eng-almeida: It doesn't. `mutateAsync` it's just another promise so you need to handle the loading state on your own.

remotecom: got it, thanks

gabrielseco:
let's merge this one

PR #42, opened 2025-03-24
feat(cost-calculator) - use premium benefits

gabrielseco:
Use premium benefits


src/flows/CostCalculator/hooks.ts
```diff
@@ -36,16 +36,23 @@ type CostCalculatorCountry = {
...
-const useCostCalculatorCountries = () => {
+const useCostCalculatorCountries = ({
+  includePremiumBenefits,
+}: {
+  includePremiumBenefits: CostCalculatorEstimationOptions['includePremiumBenefits'];
+}) => {
   const { client } = useClient();
   return useQuery({
-    queryKey: ['cost-calculator-countries'],
+    queryKey: ['cost-calculator-countries', includePremiumBenefits],
```
eng-almeida, line 46: Are we sure about this? Did you confirm internally that countries list changes depending on that flag?

gabrielseco: asked nuno about this, he will tackle this today

src/flows/CostCalculator/hooks.ts
```diff
@@ -130,18 +137,28 @@ export const useCostCalculatorEstimationPdf = () => {
...
+    includePremiumBenefits,
+  }: {
+    includePremiumBenefits: CostCalculatorEstimationOptions['includePremiumBenefits'];
+  },
+) => {
   const { client } = useClient();
 
   return useQuery({
-    queryKey: ['cost-calculator-region-fields', region],
+    queryKey: ['cost-calculator-region-fields', region, includePremiumBenefits],
```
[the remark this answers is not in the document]
gabrielseco: for the region we need them as this param will allow us to load the benefits field from the schema

src/client/sdk.gen.ts
```diff
@@ -69,6 +69,9 @@ import type {
   PostCreateCancellationData,
   PostCreateCancellationResponse,
   PostCreateCancellationError,
+  GetIndexEmploymentJobData,
```
eng-almeida, line 72: Unless there's a need for this update in this PR we should update the API SDK in a separate PR

src/flows/CostCalculator/hooks.ts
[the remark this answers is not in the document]
eng-almeida: Oh right, this one is legit 👍 Let's rename this hook to `useBenefitFields`. Then `includePremiumBenefits` will make more sense.

src/client/sdk.gen.ts
gabrielseco: yes, we need to update as the EstimationOptions were include

I don't mind to create independent MRs if you want to have a cleaner MR.

src/flows/CostCalculator/hooks.ts
```diff
@@ -176,13 +194,19 @@ export const useCostCalculator = (
...
+  const { data: countries } = useCostCalculatorCountries({
+    includePremiumBenefits: estimationOptions.includePremiumBenefits,
+  });
   const { data: currencies } = useCompanyCurrencies();
 
   const jsonSchemaRegionSlug = selectedRegion || selectedCountry?.value;
 
-  const { data: jsonSchemaRegionFields } =
-    useRegionFields(jsonSchemaRegionSlug);
+  const { data: jsonSchemaRegionFields } = useRegionFields(
```
eng-almeida, line 204: Let's rename `jsonSchemaRegionFields` to `benefitsFields`

gabrielseco: does that make sense?

jsonSchemaRegionfields can return things like benefits, fields like age or contract duration

src/flows/CostCalculator/hooks.ts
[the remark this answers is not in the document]
gabrielseco: same as before even though, we render benefits, we can render age or contract duration fields

eng-almeida (approved):

PR #43, opened 2025-03-24
feat(CostCalculator): Enable benefit submission when available

eng-almeida:
This PR adds functionality to submit selected benefits through the Cost Calculator form when they are available.

### DIFF
--- a/src/components/form/fields/FieldSetField.tsx
+++ b/src/components/form/fields/FieldSetField.tsx
@@ -22,12 +22,18 @@ type FieldWithoutOptions = FieldBase & {
 type Field = FieldWithOptions | FieldWithoutOptions;
 
 type FieldSetProps = {
-  legend: string;
+  label: string;
   name: string;
+  description: string;
   fields: Field[];
 };
 
-export function FieldSetField({ legend, name, fields }: FieldSetProps) {
+export function FieldSetField({
+  label,
+  name,
+  fields,
+  description,
+}: FieldSetProps) {
   return (
     <fieldset
       className={cn(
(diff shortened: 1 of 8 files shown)


src/components/form/fields/TextField.tsx
```diff
@@ -51,6 +51,7 @@ export function TextField({
           <FormControl>
             <Input
               {...field}
+              value={field.value ?? ''}
```
eng-almeida, line 54: Fixes a warning in the console. 
```
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component.
```

src/components/form/fields/FieldSetField.tsx
```diff
@@ -22,20 +22,32 @@ type FieldWithoutOptions = FieldBase & {
...
+  description,
+}: FieldSetProps) {
   return (
     <fieldset
       className={cn(
         'border-1 border-input p-4 rounded-xl',
         `RemoteFlows__FieldSetField__${name}`,
       )}
     >
-      <legend className="text-sm font-semibold px-2">{legend}</legend>
```
eng-almeida, line 38: Previously this would be `undefined`. It now, updates the prop from `legend` to `label` as this is what we get from `json-schema-form`.
image

eng-almeida: It also renders the `description`

gabrielseco:
One thing I noticed in my [MR](https://github.com/remoteoss/remote-flows/pull/42) is if you accidentally select Jamaica, premium benefits appear, then you select japan

The previous benefit fields from Jamaica will appear in the payload submission as undefined, after you fill japan, it seems even if we render new benefits, in the form stays.

I can investigate in my MR if you want to fix this issue

eng-almeida:
The issue you reported should be fixed. Mind to take a look again?

src/flows/CostCalculator/CostCalculator.tsx
```diff
@@ -107,6 +114,20 @@ export function CostCalculator({
     mode: 'onBlur',
   });
 
+  const { unregister, getValues } = form;
```
gabrielseco, line 117: I don't have much experience with react-hook-form.

I asked copilot about the problem and suggested to add

shouldUnregister: true

to the useForm

Did you tried that? I did and it seems to work

> 3. Using shouldUnregister Option
react-hook-form provides a shouldUnregister option when initializing the form. If set to true, it automatically unregisters fields when they are removed from the DOM.

let me know your thoughts

eng-almeida: Oh great call 🙌 It works like a charm. Thanks

remotecom:
good to go!

gabrielseco:
good to go!

PR #44, opened 2025-03-25
feat(checkbox) - Add checkbox field to the library

gabrielseco:
## Overview

Implemented the Checkbox component inside the CostCalculator for demo purposes.

Once we delete the code in the fields file about the checkbox, tests will start to pass

image

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -88,14 +88,7 @@ function App() {
 
   return (
     <div className="cost-calculator__container">
-      <RemoteFlows
-        theme={{
-          colors: {
-            checkboxBackground: 'blue',
-          },
-        }}
-        auth={() => fetchToken()}
-      >
+      <RemoteFlows auth={() => fetchToken()}>
         <CostCalculatorForm />
       </RemoteFlows>
     </div>
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -51,12 +51,7 @@ function CostCalculatorForm() {
     <>
       <CostCalculator
         estimationOptions={estimationOptions}
-        defaultValues={{
-          countryRegionSlug: 'bf098ccf-7457-4556-b2a8-80c48f67cca4',
-          currencySlug: 'eur-acf7d6b5-654a-449f-873f-aca61a280eba',
-          salary: '50000',
-        }}
-        params={{
+        options={{
           disclaimer: {
             label: 'Remote Disclaimer',
           },
@@ -90,16 +85,7 @@ function App() {
   };
 
   return (
-    <RemoteFlows
-      theme={{
-        colors: {
-          primaryBackground: 'blue',
-          accentBackground: 'green',
-          accentForeground: 'red',
-        },
-      }}
-      auth={() => fetchToken()}
-    >
+    <RemoteFlows auth={() => fetchToken()}>
       <CostCalculatorForm />
     </RemoteFlows>
   );
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -88,7 +88,14 @@ function App() {
 
   return (
     <div className="cost-calculator__container">
-      <RemoteFlows auth={() => fetchToken()}>
+      <RemoteFlows
+        theme={{
+          colors: {
+            checkboxBackground: 'blue',
+          },
+        }}
+        auth={() => fetchToken()}
+      >
         <CostCalculatorForm />
       </RemoteFlows>
     </div>
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -63,7 +63,9 @@ function CostCalculatorForm() {
         }}
       />
       {estimations && (
-        <CostCalculatorResults employmentData={estimations.data} />
+        <div className="cost-calculator__results">
+          <CostCalculatorResults employmentData={estimations.data} />
+        </div>
       )}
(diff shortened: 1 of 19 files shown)


src/flows/CostCalculator/fields.ts
```diff
@@ -48,4 +48,16 @@ export const fields: Field[] = [
...
+    inputType: 'checkbox',
+    isVisible: true,
+    label: 'Accept Terms',
+    name: 'terms',
+    required: true,
+    schema: boolean()
+      .oneOf([true], 'You must accept the terms and conditions')
+      .required('You must accept the terms and conditions'),
+    type: 'integer', // TODO: what type goes here?
+  },
```
eng-almeida, line 62: Why are we adding a accept terms checkbox to the cost calculator? Is it because of your comment in the description? I suppose then that this will be removed

src/components/form/fields/TextField.tsx
```diff
@@ -56,7 +56,7 @@ export function TextField({
                 onChange?.(event);
               }}
               {...typeAttrs}
-              className="TextField__Input"
+              className="RemoteFlows__TextField__Input"
```
eng-almeida, line 59: Thank you 🙏

eng-almeida (commented):
Looking good but I didn't test it locally yet. Have you tried to change theme tokens and see how the checkbox looks like?

src/flows/CostCalculator/fields.ts
remotecom: demo purposes

gabrielseco: this is demo to see that everything works when using the fields

I'll remove later

src/components/ui/checkbox.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { cn } from '@/src/lib/utils';
+
+const Checkbox = React.forwardRef<
+  React.ComponentRef<typeof CheckboxPrimitive.Root>,
+  React.ComponentPropsWithoutRef<typeof CheckboxPrimitive.Root>
+>(({ className, ...props }, ref) => (
+  <CheckboxPrimitive.Root
+    ref={ref}
+    className={cn(
+      'peer h-4 w-4 shrink-0 rounded-sm border border-input ring-offset-background focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50 data-[state=checked]:bg-[var(--checkbox-background)] data-[state=checked]:text-primary',
```
gabrielseco, line 16: Hey @eng-almeida,

I am doing something wrong here?

I tried using data-[state=checked]:bg-checkbox-background and it didn't work

gabrielseco: ok, I fix it

src/styles/global.css
```diff
@@ -26,6 +26,7 @@
   --focused: var(--borderRadius, #0061ff);
   --fontBase: var(--fontSizeBase, 1rem);
   --link-button: var(--foreground, #4b5865);
+  --checkbox-background: var(--checkboxBackground, #0f1419);
```
eng-almeida, line 29: Until we have more feedback about the SDK I think we should keep the number of CSS variables minimum. In this case I would use `borderInput` color. I think it makes sense that checkboxes use the same color as the other inputs
```
--checkbox-background: var(--borderInput, #9aa6b2);
```

gabrielseco: the reason was black looked good for the background color when the checkbox is selected

image

This is how it looks with your suggestion

eng-almeida: My point is that the checkbox's background color should match its border color. I don't think we need a separate color for that, at least for now. With this change, you just need to change `borderInput` token to change all the input's color.

example/src/App.tsx
```diff
@@ -83,7 +88,14 @@ function App() {
 
   return (
     <div className="cost-calculator__container">
-      <RemoteFlows auth={() => fetchToken()}>
+      <RemoteFlows
+        theme={{
+          colors: {
+            checkboxBackground: 'blue',
```
eng-almeida, line 94: `checkboxBackground`  can be remove 👍

src/types/theme.ts
```diff
@@ -15,6 +15,7 @@ export type ThemeColors = Partial<{
   primaryForeground: string;
   accentBackground: string;
   accentForeground: string;
+  checkboxBackground: string;
```
eng-almeida, line 18: `checkboxBackground`  can be remove 👍

src/components/form/fields/CheckBoxField.tsx
```diff
@@ -0,0 +1,59 @@
...
+              <div className="flex space-x-2">
+                <Checkbox
+                  id={name}
+                  onCheckedChange={field.onChange}
+                  checked={field.value}
+                  className="RemoteFlows__CheckBox__Input"
+                />
+                <FormLabel
+                  htmlFor={name}
+                  className="RemoteFlows__CheckBox__Label mb-0"
```
eng-almeida, line 46: Let's change the order so that `RemoteFlows__CheckBox__Label` class takes precedence over `mb-0`

```suggestion
                  className="mb-0 RemoteFlows__CheckBox__Label"
```

eng-almeida (approved):

PR #45, opened 2025-03-27
feat(cost-calculator) - offer a way to reset the form fields

gabrielseco:
Offer a button to reset the form fields of the CostCalculator.

Right now the button is hardcoded we can customize in the future if needed.

https://github.com/user-attachments/assets/bb75a650-3cef-4ada-95d1-fc68eb5448e9

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,7 +377,7 @@ The `useCostCalculator` hook provides access to the underlying functionality of 
 | `isSubmitting`     | `boolean`                                                                                                          | Whether the form is currently submitting                                                                                 |
 | `isLoading`        | `boolean`                                                                                                          | Whether any required data is still loading                                                                               |
 | `onSubmit`         | `(values: CostCalculatorEstimationFormValues) => Promise<Result<CostCalculatorEstimateResponse, EstimationError>>` | Function to submit the form data to the Remote API                                                                       |
-| `resetForm`        | `Function`                                                                                                         | Function to reset the form fields                                                                                        |
+| `resetForm`        | `Function`                                                                                                         | Function that clears country and region selection state                                                                  |
 
 #### Parameters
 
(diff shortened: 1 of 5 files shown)


src/flows/CostCalculator/hooks.ts
```diff
@@ -169,10 +179,11 @@ export type EstimationError = PostCreateEstimationError | ValidationError;
  * Hook to use the cost calculator.
  */
 export const useCostCalculator = (
-  { defaultRegion, estimationOptions }: UseCostCalculatorParams = {
+  { defaultRegion, estimationOptions, resetForm }: UseCostCalculatorParams = {
```
eng-almeida, line 182: I don't think it's easy to understand how to understand `resetForm` as an option to the hook. IMO `resetForm` should be returned by the hook. This way you don't need to play with useEffects and use useQueryClient.

```js

const useCostCalculator = () => {

  function resetForm() {
    setSelectedCountry(undefined);
    setSelectedRegion(defaultRegion);
  }

  return {
    resetForm
  }
}

// usage
const { fields, resetForm } = useCostCalculator();
```

In the CostCalculator you simply call `resetForm()` function, when the button is clicked.

src/flows/CostCalculator/CostCalculator.tsx
```diff
@@ -120,6 +122,14 @@ export function CostCalculator({
     }
   };
 
+  const handleReset = () => {
+    setResetForm(true);
+    form.reset();
+    setTimeout(() => {
+      setResetForm(false);
+    }, 100);
```
eng-almeida, line 130: Why the setTimeout?

src/flows/CostCalculator/hooks.ts
gabrielseco: what I was thinking...

src/flows/CostCalculator/CostCalculator.tsx
gabrielseco: nah, it was to put it to false again

README.md
```diff
@@ -377,6 +377,7 @@ The `useCostCalculator` hook provides access to the underlying functionality of
 | `isSubmitting`     | `boolean`                                                                                                          | Whether the form is currently submitting                                                                                 |
 | `isLoading`        | `boolean`                                                                                                          | Whether any required data is still loading                                                                               |
 | `onSubmit`         | `(values: CostCalculatorEstimationFormValues) => Promise<Result<CostCalculatorEstimateResponse, EstimationError>>` | Function to submit the form data to the Remote API                                                                       |
+| `resetForm`        | `Function`                                                                                                         | Function to reset the form fields                                                                                        |
```
eng-almeida, line 380: I think we should be more specific. It should be something like, `function that clears country and region selection state`

eng-almeida (approved):

PR #48, opened 2025-03-28
feat(Datepicker): Add Datepicker field

eng-almeida:
This PR adds support for the Datepicker field in forms.


eng-almeida:
@gabrielseco we need this field component but I wonder how useful this is going to be to partners. They might be using a different datepicker and therefore providing a different experience. Sure consumer apps can configure colors and spaces but that's it. While on input's and select's the experience is more a less the same and can you can change everything you need with CSS, in this case I don't think that will be possible.

gabrielseco:
This MR only covers the DateField component

Are you going to do another MR to integrate into the `JSONSchemaFormFields`?

This how it looks for me in my local

image

Shouldn't the field go to the end?

src/components/ui/calendar.tsx
```diff
@@ -0,0 +1,73 @@
...
+  classNames,
+  showOutsideDays = true,
+  ...props
+}: React.ComponentProps<typeof DayPicker>) {
+  return (
+    <DayPicker
+      showOutsideDays={showOutsideDays}
+      className={cn('p-3', className)}
+      classNames={{
+        months: 'flex flex-col sm:flex-row gap-2',
```
gabrielseco, line 19: we could add a class here if we want more customization to the calendar component right?

We could do it in the future when we have partner feedback

eng-almeida:
@gabrielseco 

> Are you going to do another MR to integrate into the JSONSchemaFormFields?

I can do it in this PR. I've just pushed the changes

> Shouldn't the field go to the end?

Good point 👍 Fixed as well

src/components/ui/calendar.tsx
eng-almeida: > We could do it in the future when we have partner feedback

Yes, I think it's preferable to wait for feedback and iterate afterwards.

gabrielseco (approved):
good to go!

PR #50, opened 2025-03-31
feat(textarea) - create textarea

gabrielseco:
Create textarea component, it's already done in a way that we can use it in the JSONSchemaForm

image
image

The count appears when we use the maxLength property

### DIFF
--- a/src/components/form/fields/TextAreaField.tsx
+++ b/src/components/form/fields/TextAreaField.tsx
@@ -1,0 +1,78 @@
+import * as React from 'react';
+
+import { useFormContext } from 'react-hook-form';
+import {
+  FormControl,
+  FormDescription,
+  FormField,
+  FormItem,
+  FormLabel,
+  FormMessage,
+} from '../../ui/form';
+import { Textarea } from '../../ui/textarea';
+import { cn } from '@/src/lib/utils';
+
+type TextAreaFieldProps = {
+  label: string;
+  description?: string;
+  name: string;
+  maxLength?: number;
+  onChange?: (event: React.ChangeEvent<HTMLTextAreaElement>) => void;
+};
+
+export function TextAreaField({
+  name,
+  description,
+  label,
+  onChange,
+  maxLength,
+}: TextAreaFieldProps) {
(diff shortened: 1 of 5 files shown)


src/components/form/fields/TextAreaField.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    'border-red-500 focus-visible:ring-red-500',
+                  'RemoteFlows__TextArea__Input',
+                )}
+              />
+            </FormControl>
+            {description && (
+              <div className="flex items-center justify-between">
+                <FormDescription className="RemoteFlows__TextArea__Description">
+                  {description}
+                </FormDescription>
```
eng-almeida, line 62: Since description is optional is better if we render FormDescription conditionally:

```jsx
{ description ? 
    <FormDescription className="RemoteFlows__TextArea__Description">
        {description}
    </FormDescription>
  :
  null
}
```

src/components/form/fields/TextAreaField.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export function TextAreaField({
+  name,
+  description,
+  label,
+  onChange,
+  maxLength,
+}: TextAreaFieldProps) {
+  const valueMaxLength = maxLength ?? null;
```
eng-almeida, line 30: I don't think this is necessary, right? The nullish coalescing operator provides a default value if the expression on its left-hand side is either null or undefined and on this case the default value it's null, which I don't think makes to much sense. `maxLength` is optional, so maxLength can be undefined (or null if for some reason the API responds with a null value) or a number. If it's 0 (which doesn't make sense) or undefined we don't show the number of characters, so:

```jsx
{maxLength && (
  <span className="text-sm RemoteFlows__TextArea__MaxLength">
    {valueLength}/{maxLength}
  </span>
)}
```
should be enough.

gabrielseco: did the necessary changes.

And check how things look when maxLength is used and description it isn't

image

eng-almeida (approved):

PR #55, opened 2025-04-03
feat(file-upload) - create a simple file upload

gabrielseco:
Create simple file upload, we can generate it from the fields file to render in the page.

The file can be multiple or not and we can remove the files that have been selected

https://github.com/user-attachments/assets/72be2c5e-988c-4e65-ab23-bd8af7526509

### DIFF
--- a/src/components/form/fields/FileUploadField.tsx
+++ b/src/components/form/fields/FileUploadField.tsx
@@ -36,6 +36,13 @@ export function FileUploadField({
             <FormLabel className="RemoteFlows__FileUpload__Label">
               {label}
             </FormLabel>
+            {description && (
+              <div className="flex items-center justify-between">
+                <FormDescription className="RemoteFlows__FileUpload__Description">
+                  {description}
+                </FormDescription>
+              </div>
+            )}
             <FormControl>
               <FileUploader
                 {...field}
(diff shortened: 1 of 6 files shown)


src/components/form/fields/FileUploadField.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <FormField
+      control={control}
+      name={name}
+      render={({ field, fieldState }) => {
+        return (
+          <FormItem className={`RemoteFlows__FileUpload__Item__${name}`}>
+            <FormLabel className="RemoteFlows__FileUpload__Label">
+              {label}
+            </FormLabel>
+            {description && (
```
eng-almeida, line 41: I think we should be consistent with the other fields and move description bellow, right before the error message.

eng-almeida (commented):
Looking good. I left only a comment and after it's addressed should be good to go.

src/components/form/fields/FileUploadField.tsx
remotecom: okay, I was intentional to mirror a bit ours, I don't mind moving it and then iterate with feedback.

eng-almeida: I see! I don't think we should consider about Remote UI for this project. IMMO this UI should be agnostic, consistent and flexible.

eng-almeida (approved):

PR #56, opened 2025-04-07
feat(cost-calculator) - use new flow component design

gabrielseco:
I did a first iteration on how the CostCalculatorFlow should work.

Tests added, docs added and old components have been removed


src/components/form/fields/SelectField.tsx
```diff
@@ -38,54 +38,55 @@ export function SelectField({
...
+      render={({ field, fieldState }) => {
+        return (
+          <FormItem className={`RemoteFlows__SelectField__Item__${name}`}>
+            <FormLabel className="RemoteFlows__SelectField__Label">
+              {label}
+            </FormLabel>
+            <FormControl>
+              <div className="relative">
+                <Select
+                  value={field.value || ''}
```
eng-almeida, line 55: Is this necessary? Did you find any bug relate to this? 

Are there any other changes in this component? It's a bit hard checking the diff

eng-almeida (commented):
Looking great! 

My only observations are about `Disclaimer` component but can be addressed in a separate PR if you prefer:
- Rename the file to `CostCalculatorDisclaimer.tsx` and move it next to `CostCalculator.tsx`
- delete `useCostCalculatorDisclaimer`. I don't think there's a need for the hook neither save the content in cache. IMO we can simply read the content from a variable.

src/components/form/fields/SelectField.tsx
gabrielseco: yes, I had a bug when I was resetting the form, it wasn't resetting the country and the currency when it has been selected.

But I didn't understand the problem well as the field.value that I was getting was ''

eng-almeida (approved):

PR #71, opened 2025-04-10
feat(custom components): Support of custom components for the remaining fields

eng-almeida:
Follow up to https://github.com/remoteoss/remote-flows/pull/68 .
Adds support to custom components for remaining fields.

### DIFF
--- a/src/components/form/fields/CheckBoxField.tsx
+++ b/src/components/form/fields/CheckBoxField.tsx
@@ -7,15 +7,17 @@ import {
   FormLabel,
   FormMessage,
 } from '@/src/components/ui/form';
+import { useFormFields } from '@/src/context';
 import { cn } from '@/src/lib/utils';
+import { JSFField } from '@/src/types/remoteFlows';
+import { CheckedState } from '@radix-ui/react-checkbox';
 import * as React from 'react';
 import { useFormContext } from 'react-hook-form';
 
-type CheckBoxFieldProps = {
-  name: string;
-  label: string;
-  description?: string;
-  defaultValue?: string;
+export type CheckBoxFieldProps = JSFField & {
+  onChange?: (
+    value: CheckedState | React.ChangeEvent<HTMLInputElement>,
+  ) => void;
 };
 
 export function CheckBoxField({
@@ -23,7 +25,10 @@ export function CheckBoxField({
   defaultValue,
   description,
   label,
+  onChange,
+  ...rest
 }: CheckBoxFieldProps) {
+  const { components } = useFormFields();
   const { control } = useFormContext();
   return (
     <FormField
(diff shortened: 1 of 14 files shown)


src/types/remoteFlows.ts
```diff
@@ -25,8 +25,11 @@ export type JSFField = {
   schema: AnySchema;
   scopedJsonSchema: Record<string, unknown>;
   type: string;
-  options?: Array<{ value: string; label: string }>;
+  options?: Array<{ value: string; label: string; description?: string }>;
```
gabrielseco, line 28: the description in options in which component we need it?

src/components/form/fields/CheckBoxField.tsx
```diff
@@ -7,37 +7,68 @@ import {
...
+            description,
+            label,
+            defaultValue,
+            ...rest,
+          };
+          return (
+            <CustomCheckboxField
+              field={{
+                ...field,
+                onChange: (value: React.ChangeEvent<HTMLInputElement>) => {
```
gabrielseco, line 52: nitpick: shouldn't be this event instead of value

src/components/form/fields/CheckBoxField.tsx
```diff
@@ -7,37 +7,68 @@ import {
...
+          );
+        }
         return (
           <FormItem className={cn(`RemoteFlows__CheckBoxField__Item__${name}`)}>
             <FormControl>
               <div className="flex space-x-2">
                 <Checkbox
                   id={name}
-                  onCheckedChange={field.onChange}
+                  onCheckedChange={(event: CheckedState) => {
```
gabrielseco, line 68: and this one value instead of event

src/components/form/fields/DatePickerField.tsx
```diff
@@ -17,82 +17,117 @@ import {
...
 import { format } from 'date-fns';
 
-type DatePickerFieldProps = {
-  description?: string;
-  label: string;
-  name: string;
-  minDate?: string;
+export type DatePickerFieldProps = JSFField & {
+  onChange?: (
+    value: Date | string | React.ChangeEvent<HTMLElement> | undefined,
```
gabrielseco, line 28: I played a bit and it seem the string type wasn't needed

src/components/form/fields/DatePickerField.tsx
```diff
@@ -17,82 +17,117 @@ import {
...
+            name,
+            minDate,
+            onChange,
+            ...rest,
+          };
+          return (
+            <CustomDatePickerField
+              field={{
+                ...field,
+                onChange: (value: React.ChangeEvent<HTMLElement>) => {
```
gabrielseco, line 61: Nitpick: shouldn't be this event?

gabrielseco (commented):
added some comments, I am switching to testing it

gabrielseco:
I was testing this with the contract amendment form

Will be valuable to them to control the render of the force value component?

That way they control the full render of the form

image

I tested the other stuff with code generated by copilot and it's working great

src/types/remoteFlows.ts
eng-almeida: The options for experience_level in contract amendments can have a description.

image

gabrielseco: Didn't see that makes sense

src/components/form/fields/CheckBoxField.tsx
eng-almeida: hmm in fact `value` should probably be `any` (like onChange from field) because we don't know how the implementation looks like on the consumer side. wdyt?

gabrielseco: I don't mind

gabrielseco (approved):

PR #73, opened 2025-04-11
feat(termination) - first approximation for the termination flow

gabrielseco:
I want to merge part of the work, what I focus was on building a json schema based on our offboarding flow, submit it to the API and check that everything went smoothly.

There are some demo files in the Example folder that we can delete before merging this to main.

Things that are still in progress: 

- [ ] email type supported in the library
- [ ] validate data in the submit function?
- [ ] risk_assessment_reasons must be an array and render a group of checkboxes
- [ ] acknowledge_termination_procedure must render the username that's doing the termination
- [ ] Steps
- [ ] Some texts that we render in the form in production, they work like a section
- [ ] Show different feedback messages in the form
- [ ] upload files to their endpoints
- [ ] retrieve timeoff data

https://github.com/user-attachments/assets/b01cb5a3-4c2d-4a43-8de3-7d6efc653242


example/src/Termination.tsx
```diff
@@ -0,0 +1,47 @@
...
+      .catch((error) => {
+        console.error({ error });
+        throw error;
+      });
+  };
+  return (
+    <RemoteFlows auth={() => fetchToken()}>
+      <div className="cost-calculator__container">
+        <TerminationFlow
+          employmentId="7df92706-59ef-44a1-91f6-a275b9149994"
```
gabrielseco, line 26: as I did the termination you probably need a different employment :/

src/flows/Termination/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+      );
+
+      const {
+        customer_informed_employee: customerInformedEmployee,
+        risk_assessment_reasons: riskAssessmentReasons,
+      } = parsedValues;
+
+      const employeeAwareness = customerInformedEmployee
+        ? {
+            employee_awareness: {
```
gabrielremote, line 102: I wonder if in the json schema you can use dot notation to avoid doing this.

Note for me for tuesday

src/flows/Termination/TerminationFlow.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export const TerminationFlow = ({
+  employmentId,
+  render,
+  options,
+}: TerminationFlowProps) => {
+  const terminationBag = useTermination({ employmentId, options });
+  if (terminationBag.isLoading) {
+    return <>{render({ terminationBag })}</>;
+  }
```
eng-almeida, line 65: I had a bug yesterday related to this. We can't have the conditional rendering here for the same component, ie the component inside the `render` prop, because it might break the rules of hooks. For example, if we had a `useState` to component inside the render prop, it will break the rules of hooks.
We can remove this. I thought we would need to have the initial values first before creating the form but it keeps working 🙂

src/flows/Termination/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (!employmentId) {
+      throw new Error('Employment id is missing');
+    }
+
+    if (terminationHeadlessForm) {
+      const parsedValues = parseJSFToValidate(
+        values,
+        terminationHeadlessForm.fields,
+        {
+          isPartialValidation: false,
```
eng-almeida, line 91: We keep passing `isPartialValidation: false` in every place we use parseJSFToValidate. We should probably change isPartialValidation default value to false.

src/flows/Termination/TerminationForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+      }
+    });
+    return () => subscription?.unsubscribe();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const handleSubmit = async (values: TerminationFormValues) => {
+    const terminationResult = await terminationBag?.onSubmit(values);
+
+    await onSubmit?.(values);
```
eng-almeida, line 37: I did this for ContractAmendment and I think it should be the rule, including Cost Calculator: onSubmit should run before `terminationBag?.onSubmit`. This is a callback for partner's to run some code before submitting to Remote. `onSuccess` and `onError` should run after  `terminationBag?.onSubmit` depending on the response we get (as it is now). Wdyt?

gabrielseco: I didn't notice, I wanted to pass the callback earlier.

I'll change it

src/flows/Termination/TerminationFlow.tsx
gabrielseco: removed

eng-almeida (approved):

PR #81, opened 2025-04-17
feat(files) - transform files to base64 to be in the desired endpoint format

gabrielseco:
I did some changes to the FileUploadField to transform the files uploaded into base64.

Right now, the termination endpoint is failing with a 500 error but I think it's a problem in the BE endpoint.

The payload is in good shape as it doesn't send a 422

### DIFF
--- a/src/components/form/fields/FileUploadField.tsx
+++ b/src/components/form/fields/FileUploadField.tsx
@@ -46,6 +46,7 @@ export function FileUploadField({
   const handleFileChange = async (
     event: React.ChangeEvent<HTMLInputElement>,
     field: ControllerRenderProps<FieldValues, string>,
+    isMultiple: boolean = false,
   ) => {
     const files = event.target.files ? Array.from(event.target.files) : [];
 
@@ -58,9 +59,14 @@ export function FileUploadField({
         };
       }),
     );
+    if (isMultiple) {
+      field.onChange(fileObjects);
+      onChange?.(fileObjects);
+      return;
+    }
 
-    field.onChange(fileObjects);
-    onChange?.(fileObjects);
+    field.onChange(fileObjects[0]);
+    onChange?.(fileObjects[0]);
   };
   return (
     <FormField
@@ -99,7 +105,7 @@ export function FileUploadField({
               <FileUploader
                 {...field}
                 onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
-                  handleFileChange(event, field)
+                  handleFileChange(event, field, multiple)
                 }
                 multiple={multiple}
                 className={cn('RemoteFlows__FileUpload__Input')}
--- a/src/components/form/fields/FileUploadField.tsx
+++ b/src/components/form/fields/FileUploadField.tsx
@@ -4,11 +4,7 @@ import * as React from 'react';
 import { useFormFields } from '@/src/context';
 import { cn } from '@/src/lib/utils';
 import { JSFField } from '@/src/types/remoteFlows';
-import {
-  ControllerRenderProps,
-  FieldValues,
-  useFormContext,
-} from 'react-hook-form';
+import { useFormContext } from 'react-hook-form';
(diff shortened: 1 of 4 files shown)


src/components/form/fields/FileUploadField.tsx
```diff
@@ -48,10 +85,9 @@ export function FileUploadField({
             <CustomFileUploadField
               field={{
                 ...field,
-                onChange: (value: any) => {
-                  field.onChange(value);
-                  onChange?.(value);
-                },
+                value: null,
+                onChange: (value: any) =>
+                  handleFileChange(value, field, multiple),
```
gabrielseco, line 90: I tested this with a custom file upload, we need to send null to avoid setting an array or an object into the input element

```tsx
components={{
        file: ({ field, fieldState, fieldData }) => {
          return (
            <>
              <label className="text-gray-700">{fieldData.label}</label>
              <input
                {...field}
                multiple={fieldData.multiple}
                type="file"
                onChange={(ev) => field.onChange(ev)}
              />
              {fieldState.error && (
                <span className="text-red-500">{fieldState.error.message}</span>
              )}
            </>
          );
        },
      }}
```

src/components/form/utils.ts
```diff
@@ -92,9 +92,22 @@ export function convertFromCents(amount?: number | string | null) {
 const trimStringValues = (values: Record<string, any>) =>
   Object.entries(values || {}).reduce<Record<string, any>>(
     (result, [key, value]) => {
-      if (typeof value === 'object') {
-        result[key] = trimStringValues(value || {});
-      } else result[key] = typeof value === 'string' ? value.trim() : value;
+      if (Array.isArray(value)) {
```
gabrielseco, line 95: I had to do some changes here, for example the termination_reason_files expects an array of items that contains content and name properties

We were not handling array correctly and it got transformed into a object instead of an array and caused the API to send a 422

src/components/form/fields/FileUploadField.tsx
eng-almeida, line 90: I think we can simplify `handleFileChange` function signature and only pass it the value (with this change the function should be renamed to `convertFilesToBase64`)

```suggestion
                onChange: async (value: any) => {
                  const fileUploadValue = await convertFilesToBase64(value);
                  field.onChange(fileUploadValue);
                  onChange?.(fileUploadValue); 
                }
```

src/components/form/fields/tests/FileUploadField.test.tsx
```diff
@@ -58,14 +58,15 @@ describe('FileUploadField Component', () => {
...
+  it('handles file upload change correctly', async () => {
     renderWithFormContext({ ...defaultProps, onChange: mockOnChange });
 
     const file = new File(['test'], 'test.txt', { type: 'text/plain' });
     const fileInput = screen.getByLabelText('File upload');
     fireEvent.change(fileInput, { target: { files: [file] } });
-
-    expect(mockOnChange).toHaveBeenCalledTimes(1);
+    await waitFor(() => {
+      expect(mockOnChange).toHaveBeenCalledTimes(1);
```
eng-almeida, line 68:
```suggestion
      expect(mockOnChange).toHaveBeenCalledOnce();
```

eng-almeida (commented):
Left a couple of comments but overall it's looking good 💪

src/components/form/fields/FileUploadField.tsx
```diff
@@ -59,14 +53,12 @@ export function FileUploadField({
         };
       }),
     );
-    if (isMultiple) {
-      field.onChange(fileObjects);
-      onChange?.(fileObjects);
-      return;
+    if (multiple) {
+      return fileObjects;
```
eng-almeida, line 57: Perhaps we could simply return fileObjects. Consumers can than check whether the array has multiple files or not. Wdyt? I would also rename `fileObjects` to `base64Files`.

eng-almeida (approved):

PR #89, opened 2025-04-23
feat(checkboxes) - render multiple checkboxes

gabrielseco:
Did some changes to support multiple checkboxes when using json schemas.

I did the changes using the Checkbox as it will allow us to select multiple options and we do not longer need to parse the values onSubmit as the values will be stored in an array

https://github.com/user-attachments/assets/f4af23cb-dd36-47c3-b89f-d618bbc704b1

Things that I didn't do in this PR:

* Test default values of the checkboxes

### DIFF
--- a/src/components/form/fields/CheckBoxField.tsx
+++ b/src/components/form/fields/CheckBoxField.tsx
@@ -13,7 +13,11 @@ import { cn } from '@/src/lib/utils';
 import { JSFField } from '@/src/types/remoteFlows';
 import { CheckedState } from '@radix-ui/react-checkbox';
 import * as React from 'react';
-import { useFormContext } from 'react-hook-form';
+import {
+  ControllerRenderProps,
+  FieldValues,
+  useFormContext,
+} from 'react-hook-form';
 
 export type CheckBoxFieldProps = JSFField & {
   onChange?: (value: any) => void;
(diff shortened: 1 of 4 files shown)


src/components/form/fields/CheckBoxField.tsx
```diff
@@ -42,13 +67,19 @@ export function CheckBoxField({
...
+            options,
             ...rest,
           };
           return (
             <CustomCheckboxField
               field={{
                 ...field,
-                onChange: (value: any) => {
+                onChange: (value: any, optionValue: any) => {
+                  if (multiple) {
```
gabrielseco, line 79: my code for supporting custom checkboxes

```tsx
<RemoteFlows
      components={{
        checkbox: ({ field, fieldState, fieldData }) => {
          return (
            <div>
              <label>{fieldData.label}</label>
              {fieldData?.options?.map((option) => (
                <div key={option.value} className="flex space-x-2">
                  <input
                    type="checkbox"
                    id={option.value}
                    onChange={(event) => {
                      field.onChange(event.target.checked, option.value);
                    }}
                    checked={field.value?.includes(option.value)}
                    className="RemoteFlows__CheckBox__Input"
                  />
                  <label
                    htmlFor={option.value}
                    className="mb-0 RemoteFlows__CheckBox__Label"
                  >
                    {option.label}
                  </label>
                </div>
              ))}
              <div className="error-message text-destructive">
                {fieldState.error && <span>{fieldState.error.message}</span>}
              </div>
            </div>
          );
        },
      }}
      auth={() => fetchToken()}
    >
```

eng-almeida (approved):

PR #90, opened 2025-04-24
feat(termination) - add feedback messages to the termination

gabrielseco:
Add feedback messages to some termination schema fields

image

### DIFF
--- a/src/flows/Termination/jsonSchema.ts
+++ b/src/flows/Termination/jsonSchema.ts
@@ -103,6 +103,12 @@ export const jsonSchema = {
           'x-jsf-presentation': {
             direction: 'column',
             inputType: 'radio',
+            statement: {
+              title:
+                'Please do not inform the employee of their termination until we review your request for legal risks. When we approve your request, you can inform the employee and we’ll take it from there.',
+              inputType: 'statement',
+              severity: 'warning',
+            },
           },
         },
         customer_informed_employee: {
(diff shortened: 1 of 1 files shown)


src/flows/Termination/jsonSchema.ts
```diff
@@ -153,6 +159,12 @@ export const jsonSchema = {
           type: 'string',
           'x-jsf-presentation': {
             inputType: 'email',
+            statement: {
```
gabrielseco, line 162: I'd like that this message appeared after the user typed the field but I don't know if this is possible

eng-almeida: Weird, this is working on contract amendments or are you referring to something different? Check the Loom

https://www.loom.com/share/ee59003bf50d476cba67ca96425b570f

eng-almeida: I think the problem is in your schema. If you add this to the `allOf` section, I think it will make it slightly better:

```json
    {
      "if": {
        "properties": {
          "personal_email": {
            "pattern": "^.*$"
          }
        }
      },
      "then": {
        "properties": {
          "personal_email": {
            "x-jsf-presentation": {
              "statement": {
                "description": "Please make sure to add the employee personal email so we can get in contact with the employee after the termination process takes effect.",
                "inputType": "statement",
                "severity": "warning"
              }
            }
          }
        }
      }
    }
```

gabrielseco: does this code mean

When there's any text in personal_email, we make the statement appear?

Because I am seeing something different

https://github.com/user-attachments/assets/92d8edb0-8bd3-4114-97da-dbce814ae024

eng-almeida: No, it's the other way around. It shows by default the statement and it gets hidden when personal email has a value.

gabrielseco: mmm, well when it has a valid email, it still shows the statement unless I am doing something wrong.

Let's leave it fixed for now

eng-almeida (approved):

PR #91, opened 2025-04-24
feat(proxy): Add proxy support to useAuth and related components

eng-almeida:
Currently, the SDK makes direct API calls to Remote API from the frontend. This is problematic for endpoints that require client tokens, as these tokens would be exposed in the frontend code.

This PR allows setting a proxy server in the SDK to avoid this problem.


gabrielseco (approved):
Nicely done!

gabrielremote (approved):

PR #96, opened 2025-04-25
feat(termination) - time off component

gabrielseco:
I created a timeoff component that we can use in the termination flow.

image


example/src/Termination.tsx
```diff
@@ -24,20 +19,23 @@ export const Termination = () => {
       <div className="cost-calculator__container">
         <TerminationFlow
           employmentId="7df92706-59ef-44a1-91f6-a275b9149994"
-          render={({ terminationBag }) => {
+          render={({ terminationBag, components }) => {
             if (terminationBag.isLoading) {
               return <div>Loading...</div>;
             }
 
+            const { Form, SubmitButton, TimeOff } = components;
```
gabrielseco, line 27: I changed a bit the code to emit the components in the renderProp

example/src/Termination.tsx
```diff
@@ -24,20 +19,29 @@ export const Termination = () => {
...
+
             return (
               <>
-                <TerminationForm
+                <TimeOff>
+                  {({ username }) => (
+                    <>
```
gabrielseco, line 33: I use a renderProp to emit the username in case the partner wants to build a link that opens a modal

src/data/hooks.ts
```diff
@@ -0,0 +1,112 @@
+import {
```
gabrielseco, line 1: I created this file for sharing hooks, I had to use the useEmployment in the termination flow

src/flows/Termination/tests/TimeOff.test.tsx
```diff
@@ -0,0 +1,95 @@
+import { FormFieldsProvider } from '@/src/RemoteFlowsProvider';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import React from 'react';
+import { PropsWithChildren } from 'react';
+import { beforeEach, describe, it, vi } from 'vitest';
+import { http, HttpResponse } from 'msw';
+import { employment } from '@/src/flows/ContractAmendment/tests/fixtures'; // TODO: Move this to a shared folder????
```
gabrielseco, line 7: I have doubts if we should move the employment to a shared fixture file

src/flows/Termination/TimeOff.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const daysLiteral = days > 1 || days === 0 ? 'days' : 'day';
+
+  return (
+    <div className="RemoteFlows__TimeOff">
+      <p className="RemoteFlows__TimeOff__Title">
+        We have recorded {days} {daysLiteral} of paid time off for {username}
+      </p>
+      {children?.({ employment: employmentData })}
+    </div>
+  );
```
eng-almeida, line 38: I don't think the SDK should have a hardcode copy - `We have recorded {days} {daysLiteral} of paid time off for {username}`. If Bamboo what's to use a different copy they would need a new SDK version. If i18n is required this wouldn't work.  We should simply pass the timeoff and employment data back to the render prop and let the consumer render what they need.

Another thing is that we should follow what we are already doing in other components and use the `render` prop instead of `children`.

src/data/hooks.ts
```diff
@@ -0,0 +1,112 @@
...
+ * }
+ *
+ * return (
+ *   <div>
+ *     {data.map((timeOff) => (
+ *       <div key={timeOff.id}>{timeOff.reason}</div>
+ *     ))}
+ *   </div>
+ * );
+ */
```
eng-almeida, line 86: In this particular case, I think the examples, for these functions are too much and don't add anything new IMO, so I would remove it.

src/data/hooks.ts
eng-almeida: I would rename this folder to `src/common/` as it might be useful for other stuff.

eng-almeida (approved):

src/data/hooks.ts
gabrielseco: done

src/flows/Termination/tests/TimeOff.test.tsx
gabrielseco: moved it to common

src/flows/Termination/TimeOff.tsx
gabrielseco: I have changed it

eng-almeida (approved):

PR #98, opened 2025-04-28
feat(termination-flow) - add tests

gabrielseco:
Add tests that check that the termination flow correctly submits. I created some helpers to help us fill forms in the future


src/flows/Termination/tests/TerminationFlow.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+    const day = date.getDay();
+
+    return {
+      year,
+      month,
+      day,
+    };
+  }
+
+  async function fillStep1(
```
eng-almeida, line 105: nit: I would rename this function to `fillEmployeeCommunication`. I think it's more explicit! Same goes to the other functions bellow

src/common/testHelpers.ts
```diff
@@ -0,0 +1,61 @@
...
+  const checkbox = screen.getByRole('checkbox', {
+    name: new RegExp(checkboxName, 'i'),
+  });
+  expect(checkbox).toBeInTheDocument();
+  await user.click(checkbox);
+}
+
+export async function selectDayInCalendar(day: string) {
+  const user = userEvent.setup();
+  const datePickerButton = screen.getByTestId('date-picker-button');
```
eng-almeida, line 47: non blocker (can be done in a separate PR): this helper won't work as expected if there are multiple date pickers in a form

eng-almeida (approved):
Nice work! I left 2 non blocking comments

src/common/testHelpers.ts
gabrielseco: yes, I didn't spend much time on it, just copied from your code and move on. 

I will track this on linear

PR #99, opened 2025-04-29
feat(termination) - inject components into description

gabrielseco:
Inject component into description, if we inline functions we need a way to render the component, I think most of the times inline function is what we want to add different props

image

### DIFF
--- a/example/src/Termination.tsx
+++ b/example/src/Termination.tsx
@@ -1,6 +1,21 @@
 import { TerminationFlow, RemoteFlows } from '@remoteoss/remote-flows';
 import './App.css';
 
+function PersonalEmailDescription({ onClick }: { onClick: () => void }) {
+  return (
+    <p>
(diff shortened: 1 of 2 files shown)


example/src/Termination.tsx
```diff
@@ -19,6 +34,19 @@ export const Termination = () => {
       <div className="cost-calculator__container">
         <TerminationFlow
           employmentId="7df92706-59ef-44a1-91f6-a275b9149994"
+          options={{
+            jsfModify: {
+              fields: {
+                personal_email: {
+                  description: () => (
+                    <PersonalEmailDescription
+                      onClick={() => console.log('click anchor')}
```
gabrielseco, line 43: if we didn't have any props you could just pass PersonalEmailDescription to the previous implementation and it would work

eng-almeida (approved):

PR #100, opened 2025-04-29
fix(contract amendment): refactor field values handling for form navigation

eng-almeida:
When conditional fields are changed they should be visible when navigation back to the form.

https://www.loom.com/share/c3813d17c52b4e4580a6d6d2a9bc2868


gabrielseco (approved):
Approving

PR #103, opened 2025-04-29
feat(form) - form description supports normal text, html and components

gabrielseco:
Add support for html tags, normal text and rendering components


src/components/ui/form.tsx
```diff
@@ -144,9 +144,12 @@ function FormDescription({
       data-slot="form-description"
       id={formDescriptionId}
       className={cn('text-base-color text-xs', className)}
+      {...(typeof children === 'string'
```
eng-almeida, line 147: We should prevent XSS exposure and therefore we should not accept whatever comes in the children. We can use [DomPurify](https://github.com/cure53/DOMPurify) like we do at Remote.

eng-almeida (approved):

PR #104, opened 2025-04-30
fix(termination) - fix conditional fields when you go back

gabrielseco:
If you fill now conditional fields and you go back the fields are filled and with the correct state.

Similar to https://github.com/remoteoss/remote-flows/pull/100 but with some changes


src/flows/Termination/hooks.ts
```diff
@@ -235,7 +239,8 @@ export const useTermination = ({
       if (terminationHeadlessForm) {
         const parsedValues = parseJSFToValidate(
           values,
-          entireTerminationSchema?.fields,
+          terminationHeadlessForm?.fields,
```
gabrielseco, line 242: we need isPartialValidation if not the conditional fields were hidden and not parsed

eng-almeida: Ah nice 👌

eng-almeida (approved):

PR #110, opened 2025-05-01
fix(termination) - fix 2 bugs related to radios

gabrielseco:
1 bug was going to the second step, filling the form, selecting will_challenge_termination to true and the conditional description field was not showing

2 bug is select a radio immediately that allows a conditional field to appear, it didn't happen

In the video, you can take a look at both bugs 

https://github.com/user-attachments/assets/1d8f72f6-2394-4a12-940c-1cd145b51fb2

### DIFF
--- a/src/flows/Termination/TerminationForm.tsx
+++ b/src/flows/Termination/TerminationForm.tsx
@@ -36,8 +36,12 @@ export function TerminationForm({
 
   useEffect(() => {
     const subscription = form?.watch((values) => {
-      if (Object.keys(form.formState.dirtyFields).length > 0) {
-        // TODO: for some reason isDirty doesn't work the first time we touch the form
+      const isAnyFieldDirty = Object.keys(values).some(
+        (key) =>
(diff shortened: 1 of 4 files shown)


src/flows/Termination/TerminationForm.tsx
```diff
@@ -36,8 +36,12 @@ export function TerminationForm({
...
-      if (Object.keys(form.formState.dirtyFields).length > 0) {
-        // TODO: for some reason isDirty doesn't work the first time we touch the form
+      const isAnyFieldDirty = Object.keys(values).some(
+        (key) =>
+          values[key as keyof TerminationFormValues] !==
+          terminationBag?.initialValues?.[key as keyof TerminationFormValues],
+      );
+      if (isAnyFieldDirty) {
         terminationBag?.checkFieldUpdates(values);
       }
```
dragidavid, line 46: This is great, always worth checking with forms! 👍

dragidavid (approved):
I think that looks good. Can't see anything obvious! 👌

PR #121, opened 2025-05-06
fix(contract amendment): update default form values to use stepState

eng-almeida:
Fixes a bug in the contract amendment payload that was sending an empty object. This issue was introduced when we switched to using useStepState.

### DIFF
--- a/src/flows/ContractAmendment/tests/ContractAmendmentFlow.test.tsx
+++ b/src/flows/ContractAmendment/tests/ContractAmendmentFlow.test.tsx
@@ -92,7 +92,7 @@ describe('ContractAmendmentFlow', () => {
     vi.clearAllMocks();
   });
 
-  it.only('submits the form when contract details are changed', async () => {
+  it('submits the form when contract details are changed', async () => {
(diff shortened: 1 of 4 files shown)


src/flows/ContractAmendment/tests/ContractAmendmentFlow.test.tsx
```diff
@@ -80,14 +82,17 @@ describe('ContractAmendmentFlow', () => {
...
+      }),
     );
   });
 
   afterEach(() => {
     vi.clearAllMocks();
   });
 
-  it('submits the form when contract details are changed', async () => {
+  it.only('submits the form when contract details are changed', async () => {
```
gabrielseco, line 95: remove only and good to go

gabrielseco (approved):
Good to go!

src/flows/ContractAmendment/tests/ContractAmendmentFlow.test.tsx
eng-almeida: 🤦

PR #120, opened 2025-05-05
termination divided by steps

gabrielseco:
I created some components to divide the components by steps

PR #122, opened 2025-05-06
fix(termination) - Fix termination submission

gabrielseco:
Now the emit the same values that we send to the BE

### DIFF
--- a/src/flows/Termination/TerminationForm.tsx
+++ b/src/flows/Termination/TerminationForm.tsx
@@ -3,13 +3,17 @@ import { Form } from '@/src/components/ui/form';
 import React, { useEffect } from 'react';
 import { useTerminationContext } from './context';
 import { TerminationFormValues } from '@/src/flows/Termination/types';
-import { OffboardingResponse, PostCreateOffboardingError } from '@/src/client';
+import {
+  CreateOffboardingParams,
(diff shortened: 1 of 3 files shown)


src/flows/Termination/TerminationForm.tsx
```diff
@@ -50,15 +54,7 @@ export function TerminationForm({
...
   const handleSubmit = async (values: TerminationFormValues) => {
-    const lastStep =
-      terminationBag?.stepState.currentStep.index ===
-      terminationBag?.stepState.totalSteps - 1;
-
-    if (lastStep) {
-      await onSubmit?.(values);
-    }
-
-    const terminationResult = await terminationBag?.onSubmit(values);
```
eng-almeida, line 61: You're doing this because you want to avoid running `parseJSFToValidate` again, right? I would prefer doing like I'm doing it in Contract Amendment and run `parseJSFToValidate` again, instead of having the pass the callback to `terminationBag?.onSubmit`. Wdyt?

gabrielseco: I've implemented it in #120 and I am using the parseJSFToValidate.

Probably we can close this one if we're happy with #120

gabrielseco:
Closing, implemented in #120

PR #125, opened 2025-05-07
feat(json-schema-version) - support json-schema-version param

gabrielseco:
The only json schema that supports json-schema-version in the future will be contract amendments as the cost calculator region fields is entirely dynamic

### DIFF
--- a/src/flows/ContractAmendment/ContractAmendmentFlow.tsx
+++ b/src/flows/ContractAmendment/ContractAmendmentFlow.tsx
@@ -9,7 +9,7 @@ import { ContractAmendmentParams } from './types';
 
 type TUseContractAmendment = ReturnType<typeof useContractAmendment>;
 
-export type RenderProps = {
+export type ContractAmendmentRenderProps = {
   /**
    * The contract amendment bag returned by the useContractAmendment hook.
    * This bag contains all the methods and properties needed to handle the contract amendment flow.
@@ -35,7 +35,7 @@ type ContractAmendmentFlowProps = ContractAmendmentParams & {
   render: ({
     contractAmendmentBag,
     components,
-  }: RenderProps) => React.ReactNode;
+  }: ContractAmendmentRenderProps) => React.ReactNode;
 };
 
 export function ContractAmendmentFlow({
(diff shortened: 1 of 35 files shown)


eng-almeida:
If I understood this right this only allows to select a single version for json schema, right? If that's the case I don't think this is the best approach because this isn't flexible enough for future needs. 
Say that we have a new flow using a json schema that can be versioned, eg onboarding, and you want to pick _version 1_ for the onboarding and _version 2_ for contract amendments. How would that work?

We should be able to select different json schema versions for different json schemas. We have 2 options:

**Option 1.** 

Have `jsonSchemaVersion` in `RemoteFlows` (as it is) that accepts an object with all the schemas (fully typed).

```js
jsonSchemaVersion={{
  contract_amendments: '1',
  contract_details: '2'
}}
```

**Option 2**

Have `jsonSchemaVersion` at the flows component level that accepts an object with all the schemas used in that flow.

```jsx
<ContractAmendmentFlow
   jsonSchemaVersion={{
    contract_amendments: '1',
    contract_details: '2'
  }}  
/>
```

IMO `Option 2` makes more sense, since is _closer_ to the component which makes it easier to understand what's happening in that flow (as opposed to Option 1, that is way up in the tree).

Let me know your thoughts

src/flows/ContractAmendment/ContractAmendmentFlow.tsx
```diff
@@ -42,12 +42,14 @@ export function ContractAmendmentFlow({
   employmentId,
   countryCode,
   options,
+  jsonSchemaVersion,
```
eng-almeida, line 45: I think it makes sense to move `jsonSchemaVersion` inside `options`. Wdyt?

src/flows/ContractAmendment/types.ts
```diff
@@ -6,8 +6,8 @@ export type ContractAmendmentParams = {
     jsfModify?: {
       [K in 'fields' | 'allFields' | 'create' | 'pick' | 'orderRoot']?: any;
     };
-  };
-  jsonSchemaVersion?: {
-    contract_amendments?: number;
+    jsonSchemaVersion?: {
+      contract_amendments?: number;
```
eng-almeida, line 10: double checking: could this be a string? like `2.1`

gabrielseco: the types from the openapi docs indicate a number

eng-almeida: Ok, cool

PR #133, opened 2025-05-08
feat(demo) - group all demos into the same file and allow navigation between one to another

gabrielseco:
I build a simple navigation to allow switching between the different demos, I also built a simple tab to copy the code.

I'd like to host it somewhere in the future to allow other stakeholders to check it live easily, partnerships can show it to other customers etc...

image

### DIFF
--- a/example/package-lock.json
+++ b/example/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.0.0",
       "hasInstallScript": true,
       "dependencies": {
+        "@radix-ui/react-collapsible": "1.1.2",
         "@remoteoss/remote-flows": "file:../",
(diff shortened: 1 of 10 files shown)


eng-almeida (approved):
Great stuff 👍 What do you think about adding [react-syntax-highlighter](https://github.com/react-syntax-highlighter/react-syntax-highlighter) for the code blocks?

gabrielseco:
> Great stuff 👍 What do you think about adding [react-syntax-highlighter](https://github.com/react-syntax-highlighter/react-syntax-highlighter) for the code blocks?

Let me take a look

gabrielseco:
added @eng-almeida 
image

PR #141, opened 2025-05-13
feat(basic-information) - submit employment

gabrielseco:
I put some basic stuff to submit the basic information step, there are things missing like types or better handling of all fields.

I observed some bugs related to departments for example but this could unblock stuff for other steps


src/flows/Onboarding/hooks.ts
```diff
@@ -46,30 +57,63 @@ const useJSONSchemaForm = ({
...
     });
 
-  // TODO: TBD
-  async function onSubmit() {
-    return;
+  async function onSubmit(values: Record<string, unknown>) {
+    const payload: EmploymentCreateParams = {
+      basic_information: values,
+      type: employeeType,
+      external_id: externalId,
```
eng-almeida, line 113: Do we need the external_id? What's the use case for having it?

example/src/Onboarding.tsx
```diff
@@ -41,7 +51,13 @@ export const OnboardingEOR = () => {
   return (
     <RemoteFlows auth={fetchToken}>
       <div className="cost-calculator__container">
-        <OnboardingFlow employmentId={'1234'} render={MultiStepForm} />
+        <OnboardingFlow
+          employmentId="1234"
+          countryCode="PRT"
+          employeeType="employee"
```
eng-almeida, line 57: I think we should rename this to `type`, ie same name as we have in `EmploymentCreateParams`. We just need to document the props in `OnboardingFlowProps` component.

src/flows/Onboarding/hooks.ts
```diff
@@ -17,9 +26,11 @@ type OnboardingHookProps = {
 const useJSONSchemaForm = ({
   countryCode,
   form,
+  formValues,
 }: {
   countryCode: string;
   form: string;
+  formValues: Partial<$TSFixMe>;
```
eng-almeida, line 33: Let's be consistent with the other flows and rename this to `fieldValues` with type `FieldValues | undefined`

src/flows/Onboarding/hooks.ts
gabrielseco: I asked internally

src/flows/Onboarding/hooks.ts
gabrielseco: I should do the same with the termination, used formValues in there too

src/flows/Onboarding/BasicInformationStep.tsx
```diff
@@ -2,18 +2,56 @@ import React from 'react';
...
+  /*
+   * The function is called when an error occurs during form submission.
+   */
+  onError?: (error: unknown) => void;
 };
 
-export function BasicInformationStep({ onSubmit }: BasicInformationStepProps) {
+export function BasicInformationStep({
+  onSubmit,
+  onError,
```
eng-almeida, line 37: We're missing the `onSuccess` callback 🙂

src/flows/Onboarding/BasicInformationStep.tsx
```diff
@@ -2,18 +2,56 @@ import React from 'react';
...
+}: BasicInformationStepProps) {
   const { onboardingBag } = useOnboardingContext();
-  const handleSubmit = async (values: unknown) => {
-    await onSubmit?.(onboardingBag?.parseFormValues(values) as unknown);
-    onboardingBag?.next();
+  const handleSubmit = async (payload: Record<string, unknown>) => {
```
eng-almeida, line 40: Can't we set the payload type to `BasicInformationFormPayload`? We'd need to adjust `onSubmit` in `OnboardingForm` to accept an union of types

eng-almeida (approved):

PR #148, opened 2025-05-16
feat(benefits) - add benefits step without submission

gabrielseco:
Modified the benefits step to read the fields and do the validation.

Right now the submission is skipped but next button, lets the user go to the last step.

Did this to unblock an alpha release for today


src/components/form/Form.tsx
```diff
@@ -0,0 +1,15 @@
...
+};
+
+export const useForm = ({ fields }: useFormProps) => {
+  const fieldValues = {};
+  const initialValues = getInitialValues(fields, fieldValues);
+
+  return {
+    initialValues,
+  };
+};
```
eng-almeida: I don't think we need a hook just for this. Or are you planning to add other stuff in here?

src/flows/Onboarding/hooks.ts
```diff
@@ -143,6 +150,55 @@ const useJSONSchemaForm = ({
...
+        throw new Error('Failed to fetch benefit offers schema');
+      }
+
+      return response;
+    },
+    select: ({ data }) => {
+      const { schema } = modify(
+        data.data?.schema || {},
+        options?.jsfModify || {},
+      );
```
eng-almeida, line 192: It's probably ok to do this but I would rather execute `modify` only `options?.jsfModify` is defined.

src/components/form/Form.tsx
gabrielseco: I removed it, I didn't want it to include here

eng-almeida (approved):

PR #152, opened 2025-05-16
fix(onboarding): fix initial values when navigating back

eng-almeida:
Restore field values for current step when navigating back in the Onboarding flows (applied the same strategy to ContractAmendment as well)


src/flows/Onboarding/hooks.ts
```diff
@@ -143,9 +143,13 @@ const useJSONSchemaForm = ({
       const employmentField = jsonSchemaToEmployment[form] as keyof Employment;
       const employmentFieldData = (employment?.[employmentField] ||
         {}) as Record<string, unknown>;
-      return createHeadlessForm(schema, {
+
+      console.log('before createHeadlessForm', hasFieldValues, fieldValues);
```
gabrielseco: remove the log and good to go

PR #153, opened 2025-05-16
Onboarding flow

gabrielseco:
Add onboarding flow


.changeset/tender-baths-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@remoteoss/remote-flows': minor
```
gabrielseco, line 2: I will edit the changelog, package, package-lock to set correctly the alpha in the PR that CI generates @eng-almeida

PR #158, opened 2025-05-19
feat(benefits) - handle submission and retrieve the saved values

gabrielseco:
Add benefits submission and set the correct benefit initialValues

I also exported types to make the usage easier on the partner side

https://github.com/user-attachments/assets/b9e9352f-ccba-473c-8a26-004850589cf4

### DIFF
--- a/src/flows/Onboarding/hooks.ts
+++ b/src/flows/Onboarding/hooks.ts
@@ -93,6 +93,18 @@ const useBenefitOffers = (employmentId: string | undefined) => {
         return response;
       });
     },
+    select: ({ data }) =>
+      data?.data?.reduce(
+        (acc, item) => {
+          return {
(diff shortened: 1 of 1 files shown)


src/flows/Onboarding/hooks.ts
```diff
@@ -314,6 +345,28 @@ export const useOnboarding = ({
...
+    (acc, item) => {
+      return {
+        ...acc,
+        [item.benefit_group.slug]: {
+          value: item.benefit_tier?.slug ?? '',
+        },
+      };
+    },
+    {} as Record<string, { value: string }>,
+  );
```
eng-almeida, line 358: We can move this logic to `select` in `useBenefitOffers`

eng-almeida (approved):

PR #159, opened 2025-05-19
refactor(Onboarding): integrate getInitialValues for setting up initial form values

eng-almeida:
Use the `getInitialValues` function to adjust API values to form fields, such as converting `money` fields from cents.

### DIFF
--- a/src/flows/Onboarding/hooks.ts
+++ b/src/flows/Onboarding/hooks.ts
@@ -23,7 +23,10 @@ import { useMutation, useQuery } from '@tanstack/react-query';
 import { useClient } from '@/src/context';
 import { useStepState } from '@/src/flows/useStepState';
 import { STEPS } from '@/src/flows/Onboarding/utils';
-import { parseJSFToValidate } from '@/src/components/form/utils';
+import {
+  getInitialValues,
+  parseJSFToValidate,
+} from '@/src/components/form/utils';
 import { mutationToPromise } from '@/src/lib/mutations';
 import { FieldValues } from 'react-hook-form';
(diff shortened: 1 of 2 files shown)


src/flows/Onboarding/hooks.ts
```diff
@@ -382,19 +385,25 @@ export const useOnboarding = ({
...
+      stepFields[stepState.currentStep.name as keyof typeof stepFields],
+      employment?.data?.data.employment?.basic_information || {},
+    ),
+    contract_details: getInitialValues(
+      stepFields[stepState.currentStep.name as keyof typeof stepFields],
+      employment?.data?.data.employment?.contract_details || {},
+    ),
+  };
+
+  console.log('hook initialValues', initialValues);
```
gabrielseco, line 406: we need to remove the log

gabrielseco (approved):

src/flows/Onboarding/hooks.ts
```diff
@@ -378,12 +379,38 @@ export const useOnboarding = ({
...
+      stepFields[stepState.currentStep.name as keyof typeof stepFields],
+      employment?.data?.data.employment?.contract_details || {},
+    ),
+    benefits: initialValuesBenefitOffers || {},
+  };
+
   async function onSubmit(values: FieldValues) {
     switch (stepState.currentStep.name) {
       case 'basic_information': {
+        const parsedValues = parseJSFToValidate(
```
gabrielseco, line 404: we're parsing inside the step too before we call the onSubmit function, can we change the parseFormValues returned by the hook to return always the same

gabrielseco: discard

PR #177, opened 2025-05-22
fix(ForcedValueField) - handle undefined statement gracefully

eng-almeida:
Updated ForcedValueField component to conditionally render title and description based on the presence of the statement. If the statement is undefined, it now falls back to description.

### DIFF
--- a/src/components/form/fields/ForcedValueField.tsx
+++ b/src/components/form/fields/ForcedValueField.tsx
@@ -5,7 +5,7 @@ type ForcedValueFieldProps = {
   name: string;
   value: string;
   description: string;
-  statement: {
+  statement?: {
     title: string;
     description: string;
   };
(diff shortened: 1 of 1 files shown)


src/components/form/fields/ForcedValueField.tsx
```diff
@@ -26,16 +26,25 @@ export function ForcedValueField({
 
   return (
     <div>
-      <p
```
gabrielremote, line 29: I don't know how to point to line 8, the statement should be optional right?

src/components/form/fields/ForcedValueField.tsx
```diff
@@ -26,16 +26,25 @@ export function ForcedValueField({
...
-        className={`text-sm RemoteFlows__ForcedValue__Title__${name}`}
-        dangerouslySetInnerHTML={{ __html: statement.title }}
-      />
-      <p
-        className={`text-xs RemoteFlows__ForcedValue__Description__${name}`}
-        dangerouslySetInnerHTML={{
-          __html: statement.description || description,
-        }}
-      />
+      {statement ? (
```
gabrielseco, line 29: I don't know how to point to line 8, the statement should be optional right?

PR #179, opened 2025-05-26
feat(credit-risk)  - add credit risk flows

gabrielseco:
There are some parts missing here

* Being able to create a reserve invoice the API endpoint is not ready yet

image

image


src/flows/Onboarding/hooks.ts
```diff
@@ -72,6 +74,36 @@ const useEmployment = (employmentId: string | undefined) => {
...
+        },
+      });
+
+      if (response.error || !response.data) {
+        throw new Error('Failed to fetch employment data');
+      }
+
+      return response;
+    },
+    // TODO: OpenApi types are not matching, so we need to cast the response
```
eng-almeida, line 100: We should report an issue internally about this 👍

src/flows/Onboarding/hooks.ts
```diff
@@ -130,6 +162,22 @@ export const useEmploymentInvite = () => {
...
+  const { client } = useClient();
+  return useMutation({
+    mutationFn: (payload: { employment_id: string }) => {
+      return postInviteEmploymentInvitation({
+        client: client as Client,
+        headers: {
+          Authorization: ``,
+        },
+        path: payload,
+      });
```
eng-almeida, line 176: I would simply return a `Promise.resolve` to avoid confusion

src/flows/Onboarding/hooks.ts
```diff
@@ -72,6 +74,36 @@ const useEmployment = (employmentId: string | undefined) => {
   });
 };
 
+const useCompany = (companyId: string) => {
```
eng-almeida, line 77: The `hooks.ts` is getting bigger an bigger. I suggest we have in this file only the `useOnboarding` hook and move API hooks to a new file `onboarding/api.ts`. We should then have the same pattern for the all the other flows (in a separate PR). Wdyt?

src/flows/Onboarding/OnboardingCreateReserve.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const handleSubmit = async () => {
+    try {
+      await onSubmit?.();
+
+      if (!onboardingBag.employmentId) {
+        throw new Error('Employment ID is required');
+      }
+
+      const response = await createReserveMutationAsync({
```
eng-almeida, line 37: As a safety measure we should only allow running `createReserveMutationAsync` only if `creditRiskStatus === 'deposit_required'`

And the `OnboardingInvite` component should also be update with a similar verification, probably `creditRiskStatus !== 'deposit_required'`

src/flows/Onboarding/hooks.ts
gabrielseco: I did a PR in the BE, I'll let you know if I fixed the issue

src/flows/Onboarding/hooks.ts
gabrielseco: that's a good idea

gabrielseco:
I will do some changes tomorrow as I did a sync with the PM in slack cc: @eng-almeida

src/flows/Onboarding/InviteSection.tsx
```diff
@@ -0,0 +1,79 @@
...
+  );
+
+  const copyDescription = (
+    <>
+      If you're ready to invite this employee to onboard with Remote, click the
+      button below.
+    </>
+  );
+
+  return render({
```
gabrielremote, line 60: I changed the API for both components with the render prop you can render whatever you want it gives you props that were defined internally or just render the DefaultComponent

For example

```
<InviteSection
            render={({ DefaultComponent }) => {
              return (
                <>
                  <DefaultComponent title="Ready to invite employee? or Not ready yet?">
                    injected children to default component
                  </DefaultComponent>
                </>
              );
            }}
          />
```

You can override title, description, children

In the DepositRequiredSection looks like this

```
  <DepositRequiredSection
            render={({ DefaultComponent, props }) => {
              return (
                <>
                  <DefaultComponent title="Ready to invite employee? or Not ready yet?">
                    injected children to default component
                  </DefaultComponent>
                  <p>Reserve payment required to hire this employee</p>
                  <a href={props.supportLink}>What is a reserve payment</a>
                </>
              );
            }}
          />
```

src/flows/Onboarding/hooks.ts
```diff
@@ -497,7 +208,23 @@ export const useOnboarding = ({
     /**
      * Employment id passed useful to be used between components
      */
-    employmentId,
+    employmentId: internalEmploymentId,
```
gabrielseco, line 211: we need to always have the employmentId correctly set

PR #180, opened 2025-05-26
feat(useStepState): allow jumping directly to a step

eng-almeida:
Implemented logic in `useStepState` to allow jumping directly to a step if it has been filled.


gabrielseco (approved):
Looks good!

PR #182, opened 2025-05-27
feat(onboarding): Prettify form values

eng-almeida:
This PR adds a `metadata` property to the onboarding hook. In this field, we're including prettified form values and field label, so it can be used in the review step.

### DIFF
--- a/example/src/Onboarding.tsx
+++ b/example/src/Onboarding.tsx
@@ -25,38 +25,21 @@ type MultiStepFormProps = {
 };
 
 function Review({
-  values,
   meta,
 }: {
-  values: Record<string, unknown>;
-  meta: Record<string, unknown>;
+  meta: Record<string, { label: string; prettyValue: string }>;
 }) {
-  console.log('META', meta);
   return (
     <div className="onboarding-values">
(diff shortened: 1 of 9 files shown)


gabrielseco (commented):
Add one comment

gabrielseco:
I found that fieldsets aren't applied correctly

I tried for example filling the manager field and the value didn't appear in the review

example/src/Onboarding.tsx
```diff
@@ -45,14 +44,16 @@ function Review({
                 };
                 return (
                   <pre>
-                    {val.label}: {val.prettyValue}
+                    {val.label}:{' '}
+                    {value.prettyValue === true ? 'Yes' : value.prettyValue}
```
gabrielseco, line 48: `{value.prettyValue === true ? 'Yes' : value.prettyValue}`

instead of returning value.prettyValue return 'No' that way we solve the radio values where we choose a No

image

eng-almeida: Good point! Done 👍

gabrielseco (approved):

PR #186, opened 2025-05-29
fix(onboarding): fix the values passed to createHeadlessForm

eng-almeida:
We're not passing money values in cents to `createHeadlessForm` to that computed values in `x-jsf-logic`. This PR creates introduces:
- a function to get all the money fields and convert the correspondent value to cents before passing it to `createHeadlessForm`
- a small refactor to `useJSONSchemaForm` form

### DIFF
--- a/src/flows/Onboarding/hooks.ts
+++ b/src/flows/Onboarding/hooks.ts
@@ -142,13 +142,11 @@ const useJSONSchemaForm = ({
   form,
   fieldValues,
   options,
-  employment,
 }: {
   countryCode: string;
   form: JSONSchemaFormType;
   fieldValues: FieldValues;
   options?: OnboardingHookProps['options'];
-  employment?: Employment;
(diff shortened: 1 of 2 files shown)


src/flows/Onboarding/hooks.ts
```diff
@@ -186,12 +186,25 @@ const useJSONSchemaForm = ({
...
+      //   {}) as Record<string, unknown>;
+
+      const moneyFields = findFieldsByType(jsfSchema.properties || {}, 'money');
+      const moneyFieldsData = moneyFields.reduce<Record<string, number | null>>(
+        (acc, field) => {
+          acc[field] = convertToCents(fieldValues[field]);
+          return acc;
+        },
+        {},
+      );
```
eng-almeida, line 201: This is the actual fix! The remaining code changed in this file is part of the refactor.

PR #202, opened 2025-06-04
fix(tsconfig) - fix development configuration

gabrielseco:
I did some investigation before about lodash and find some inconsistencies in the tsconfig file.

We can remove a lot of React imports not used, tell tsconfig that we're using the package in a bundler + that we are not emitting commonjs


package.json
```diff
@@ -16,7 +16,8 @@
     "changeset:version": "changeset version",
     "prepare": "npm run build",
     "prerelease": "npm run ci",
-    "release": "changeset publish"
+    "release": "changeset publish",
+    "type-check": "tsc --noEmit"
```
eng-almeida, line 20: Do we need `type-check`?

eng-almeida (approved):

PR #205, opened 2025-06-04
feat(cost-calculator) - use parseJSFToValidate internally to parse the form values of the form

gabrielseco:
We're making a breaking change here:

We're changing the payload that we emit on the onSubmit function

I checked that the estimation kept working with our results component and the estimation pdf was generated correctly.

I checked with our users and they're not using onSubmit, only onSuccess and onError

The change in the onSubmit is the next

* salary field is emitted in cents and as a number
* age field is emitted as a number instead of a string

Readme and examples were updated to reflect this change


src/flows/CostCalculator/tests/CostCalculatorFlow.test.tsx
```diff
@@ -125,7 +125,7 @@ describe('CostCalculatorFlow', () => {
       expect(mockOnSubmit).toHaveBeenCalledWith({
         country: 'POL',
         currency: 'usd-1dee66d1-9c32-4ef8-93c6-6ae1ee6308c8',
-        salary: '50000',
+        salary: 5000000,
```
gabrielseco, line 128: why this test is changed?

if you take a look to the onSubmit now we call parseFormValues which uses convertToCents and transforms the quantity to a number

src/flows/CostCalculator/tests/CostCalculatorFlow.test.tsx
```diff
@@ -260,7 +260,7 @@ describe('CostCalculatorFlow', () => {
 
     await waitFor(() => {
       expect(mockOnSubmit).toHaveBeenCalledWith({
-        age: '30',
+        age: 30,
```
gabrielseco, line 263: as now we use the parseFromValues this is transformed to a number

src/flows/CostCalculator/tests/utils.test.ts
```diff
@@ -21,8 +21,8 @@ describe('buildPayload', () => {
       employments: [
         {
           region_slug: 'US',
-          annual_gross_salary: 10000000, // convertToCents(100000)
-          annual_gross_salary_in_employer_currency: 10000000,
+          annual_gross_salary: 100000,
```
gabrielseco, line 24: we no longer do the transformation to cents here

eng-almeida (approved):

PR #209, opened 2025-06-05
feat(credit-risk) - improvements to the OnboardingInvite

gabrielseco:
I did several things in this PR

* Fix the onboarding invitation, if you did the whole flow in one go and tried to invite, it didn't work
* When the creditStatus is referred we don't render the button, same as in platform
* We refetch the employment status after we invite or generate an invoice, to avoid requesting the same info twice.
* Add example component OnboardingStatuses to render different banners
* Created a ReviewStep to avoid copy & pasting

### DIFF
--- a/example/src/OnboardingAlertStatutes.tsx
+++ b/example/src/OnboardingAlertStatutes.tsx
@@ -1,0 +1,32 @@
+import { CreditRiskStatus } from '@remoteoss/remote-flows';
+
+export const OnboardingAlertStatutes = ({
+  creditRiskStatus,
+}: {
+  creditRiskStatus: CreditRiskStatus;
+}) => {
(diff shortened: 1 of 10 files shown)


example/src/OnboardingAlertStatutes.tsx
```diff
@@ -0,0 +1,32 @@
...
+        Reserve payment required to hire this employee. Check this{' '}
+        <a href="https://support.remote.com/hc/en-us/articles/12695731865229-What-is-a-reserve-payment">
+          support article
+        </a>
+      </div>
+    );
+  }
+
+  if (creditRiskStatus === 'referred') {
+    // TODO: Where does the link go at Remote?
```
gabrielseco, line 20: I'll investigate and update in another PR

PR #215, opened 2025-06-06
feat(onboarding) - refetch company after submitting each step

gabrielseco:
As we talked yesterday the company status can change until you get to the last step.

We refetch the company data every time we submit

### DIFF
--- a/src/flows/Onboarding/hooks.ts
+++ b/src/flows/Onboarding/hooks.ts
@@ -197,6 +197,8 @@ export const useOnboarding = ({
     );
 
     const parsedValues = parseFormValues(values);
+    refetchCompany();
+
     switch (stepState.currentStep.name) {
       case 'select_country': {
         setInternalCountryCode(parsedValues.country);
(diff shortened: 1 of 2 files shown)


src/flows/Onboarding/hooks.ts
```diff
@@ -216,6 +220,7 @@ export const useOnboarding = ({
           try {
             console.log('Creating employment with payload:', payload);
```
eng-almeida, line 221: Let's remove this log

src/flows/Onboarding/hooks.ts
```diff
@@ -216,6 +220,7 @@ export const useOnboarding = ({
...
             const response = await createEmploymentMutationAsync(payload);
+            await refetchCompany();
```
eng-almeida, line 223: I have 2 questions/doubts:
- can we refetchCompany at the top of the submit button? We're repeating this line in multiple places and it sounds like we can do it.
- do we need to await for refetchCompany? ie, do we care if it succeeds or not? We're not handling errors, so I guess not. As an alternative, should we run this in parallel with the other actions? 🤔

gabrielseco: * yes
* we don't care

PR #217, opened 2025-06-06
feat(onboarding) - support the different status from the credit flows

gabrielseco:
I added code both in the ReviewStep and in the OnboardingInvite to support the different uses cases of the credit flows

I will share a video internally about these changes

### DIFF
--- a/src/flows/Onboarding/tests/OnboardingInvite.test.tsx
+++ b/src/flows/Onboarding/tests/OnboardingInvite.test.tsx
@@ -217,42 +217,6 @@ describe('OnboardingInvite', () => {
     });
   });
 
-  it('should call createReserveInvoice when creditRiskStatus is deposit_required', async () => {
-    server.use(
-      http.get('*/v1/companies/:companyId', () => {
-        return HttpResponse.json({
-          ...companyResponse,
-          data: {
-            ...companyResponse.data,
-            company: {
-              ...companyResponse.data.company,
-              default_legal_entity_credit_risk_status: 'deposit_required',
-            },
-          },
-        });
-      }),
-    );
-
-    render(<OnboardingFlow {...defaultProps} />, { wrapper });
-
-    await screen.findByText(/Step: Select Country/);
-
-    const button = await screen.findByText(/Create Reserve/i);
-    expect(button).toBeInTheDocument();
-
-    fireEvent.click(button);
-
-    await waitFor(() => {
-      expect(mockSubmit).toHaveBeenCalledTimes(1);
-    });
-
-    await waitFor(() => {
-      expect(mockSuccess).toHaveBeenCalledTimes(1);
-    });
-
-    expect(mockError).not.toHaveBeenCalled();
-  });
-
   it('should call onError when creating reserve invoice fails', async () => {
     server.use(
       http.get('*/v1/companies/:companyId', () => {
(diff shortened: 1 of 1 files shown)


src/flows/Onboarding/OnboardingInvite.tsx
```diff
@@ -67,18 +93,26 @@ export function OnboardingInvite({
...
       onClick={() => {
         handleSubmit();
       }}
     >
-      {props.children === undefined
-        ? onboardingBag.creditRiskStatus === 'deposit_required'
-          ? 'Create Reserve'
-          : 'Invite Employee'
-        : props.children}
+      {render({
```
gabrielseco, line 113: it's breaking change as we don't use children anymore

src/flows/Onboarding/OnboardingInvite.tsx
```diff
@@ -37,13 +48,23 @@ export function OnboardingInvite({
...
       ) {
         const response = await createReserveInvoiceMutationAsync({
           employment_slug: onboardingBag.employmentId,
         });
         if (response.data) {
-          await onSuccess?.(response.data as SuccessResponse);
+          await onSuccess?.(
+            response.data as SuccessResponse,
+            'created_awaiting_reserve',
+          );
```
eng-almeida, line 62: We can probably improve this callback by sending an object with some meaningful names:

```ts
await onSuccess?.({
  reserveInvoiceResponse: response.data as SuccessResponse,
  employmentStatus: 'created_awaiting_reserve',
});
```
What do you think?

src/flows/Onboarding/OnboardingInvite.tsx
```diff
@@ -55,7 +76,12 @@ export function OnboardingInvite({
           employment_id: onboardingBag.employmentId,
         });
         if (response.data) {
-          await onSuccess?.(response.data as SuccessResponse);
+          await onSuccess?.(response.data as SuccessResponse, 'invited');
```
eng-almeida, line 79: Same comment here:

```
await onSuccess?.({
  inviteResponse: response.data as SuccessResponse,
  employmentStatus: 'invited',
});
```

src/flows/Onboarding/ReviewStep.tsx
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    // Default case
+    return null;
+  };
+
+  const creditRiskType = getCreditRiskType();
+
+  return render({
+    creditRiskType: creditRiskType,
```
eng-almeida, line 69: I'm struggling with this name 😅 It doesn't seem like a type, maybe creditRiskState? Not sure. Either way we need to add JSdocs to both of this properties because it's not easy to understand the difference between these 2.

src/flows/Onboarding/OnboardingInvite.tsx
```diff
@@ -1,26 +1,37 @@
...
+    status: 'invited' | 'created_awaiting_reserve',
+  ) => void | Promise<void>;
+  onError?: (error: unknown) => void;
+  onSubmit?: () => void | Promise<void>;
+  render: (props: {
+    status: 'invited' | 'created_awaiting_reserve';
+  }) => ReactNode;
+};
+
+const FINAL_EMPLOYMENT_STATUSES: Employment['status'][] = [
```
eng-almeida, line 21: Can we rename this variable? Maybe `employmentStatusList` is good enough?

src/flows/Onboarding/OnboardingInvite.tsx
gabrielseco: will it better to use one property 

data: response depends on what endpoint you use
employmentStatus

or do you want that we emit three properties

reserveInvoiceResponse
inviteResponse
employmentStatus

gabrielseco: I went with data and employmentStatus

eng-almeida (approved):

PR #219, opened 2025-06-07
feat(onboarding)  - jump to the review step

gabrielseco:
When the status in a finalized step, let's jump to the review step automatically

### DIFF
--- a/src/flows/Onboarding/hooks.ts
+++ b/src/flows/Onboarding/hooks.ts
@@ -53,7 +53,7 @@ const stepToFormSchemaMap: Record<
   review: null,
 };
 
-const statusToGoReviewStep: Employment['status'][] = [
+const reviewStepAllowedEmploymentStatus: Employment['status'][] = [
   'invited',
   'created_awaiting_reserve',
   'created_reserve_paid',
@@ -215,7 +215,7 @@ export const useOnboarding = ({
     if (
       employmentId &&
       employment &&
-      statusToGoReviewStep.includes(employment?.status) &&
+      reviewStepAllowedEmploymentStatus.includes(employment?.status) &&
       !isLoading &&
       stepState.currentStep.name !== 'review'
     ) {
--- a/src/flows/Onboarding/hooks.ts
+++ b/src/flows/Onboarding/hooks.ts
@@ -126,15 +126,12 @@ export const useOnboarding = ({
     unknown
   >;
 
-  const isOnboardingFormEnabled =
-    Boolean(
-      stepState.currentStep.name === 'basic_information' && internalCountryCode,
-    ) ||
-    Boolean(
-      (stepState.currentStep.name === 'contract_details' &&
-        internalCountryCode) ||
-        (employmentId && internalCountryCode),
-    );
+  const isOnboardingFormEnabled = Boolean(
+    internalCountryCode &&
+      (stepState.currentStep.name === 'basic_information' ||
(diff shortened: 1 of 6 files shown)


src/flows/Onboarding/hooks.ts
```diff
@@ -145,23 +172,37 @@ export const useOnboarding = ({
...
+    }
+    return {};
+  }, [
+    stepState.currentStep.name,
+    benefitOffers,
+    stepState.values,
+    fieldValues,
+  ]);
+
+  const stepFields: Record<keyof typeof STEPS, Fields> = useMemo(
```
eng-almeida, line 191: Do we need `useMemo`? Doesn't seem like a complex task to memoize.

src/flows/Onboarding/hooks.ts
```diff
@@ -132,7 +139,27 @@ export const useOnboarding = ({
...
+          ? {
+              ...stepState.values?.[stepState.currentStep.name], // Restore values for the current step
+              ...fieldValues,
+            }
+          : serverEmploymentData,
+      options: options,
+      enabled: Boolean(
+        (stepState.currentStep.name === 'contract_details' &&
+          internalCountryCode) ||
+          (employmentId && internalCountryCode),
```
eng-almeida, line 162: I'm a bit confused with this PR. It seems we're splitting `onboardingForm` in two (basic information and contract detgails)  and that introduces extra complexity and some duplicate code.

What is the reasoning?

src/flows/Onboarding/hooks.ts
```diff
@@ -178,12 +219,104 @@ export const useOnboarding = ({
...
+    if (
+      employmentId &&
+      employment &&
+      statusToGoReviewStep.includes(employment?.status) &&
+      !isLoading &&
+      stepState.currentStep.name !== 'review'
+    ) {
+      fieldsMetaRef.current = {
+        select_country: prettifyFormValues(
+          getInitialValues(stepFields['select_country'], {
```
eng-almeida, line 241: `getInitialValues` for each step is called twice, here and bellow in `setStepValues`. We can improve this and call it only once

src/flows/Onboarding/hooks.ts
gabrielseco: before we had

```
const stepFields: Record<keyof typeof STEPS, Fields> = {
    select_country: selectCountryForm?.fields || [],
    basic_information: onboardingForm?.fields || [],
    contract_details: onboardingForm?.fields || [],
    benefits: benefitOffersSchema?.fields || [],
    review: [],
  };
```


As the onboardingForm?.fields was shared we couldn't set the values correctly when using the function `getInitialValues`


Basically the contract details weren't fetched until you reach the contract details step or now when you pass it an employmentId 

Hope it helps

src/flows/Onboarding/hooks.ts
gabrielseco: I can do that

src/flows/Onboarding/hooks.ts
```diff
@@ -53,6 +53,12 @@ const stepToFormSchemaMap: Record<
   review: null,
 };
 
+const statusToGoReviewStep: Employment['status'][] = [
```
eng-almeida, line 56:
```suggestion
const reviewStepAllowedEmploymentStatus: Employment['status'][] = [
```

eng-almeida (approved):
Left a suggestion but it's looking good 👍

PR #231, opened 2025-06-11
feat(cost-calculator) - add full demo with custom inputs

gabrielseco:
I added an example on how to use custom components with the cost calculator

When I did this I found the need to export some types to have a better usage

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -893,6 +893,81 @@ export function ContractAmendment() {
   );
 }
 `,
+  costCalculatorWithReplaceableComponents: `
+  import {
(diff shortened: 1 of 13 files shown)


example/src/App.tsx
```diff
@@ -892,6 +893,81 @@ export function ContractAmendment() {
   );
 }
 `,
+  costCalculatorWithReplaceableComponents: `
```
eng-almeida, line 896: We should move code samples to a dedicated file(folder). This file is getting hard to follow.

gabrielseco: AMAZING. CHANGED!

PR #233, opened 2025-06-12
feat(onboarding) - prevent user from going back when they're in a final status

gabrielseco:
When recover an employment, who is invited, we moved the user to the review step which means that we cannot longer go back.

Besides API doesn't let do any updates to an employment in that situation

### DIFF
--- a/example/src/ReviewStep.tsx
+++ b/example/src/ReviewStep.tsx
@@ -201,6 +201,7 @@ export const ReviewStep = ({
       <button
         className="back-button"
         onClick={() => onboardingBag.goTo('basic_information')}
+        disabled={onboardingBag.isEmploymentInFinalState}
       >
         Edit Basic Information
       </button>
@@ -209,6 +210,7 @@ export const ReviewStep = ({
       <button
         className="back-button"
         onClick={() => onboardingBag.goTo('contract_details')}
+        disabled={onboardingBag.isEmploymentInFinalState}
       >
         Edit Contract Details
       </button>
@@ -218,6 +220,7 @@ export const ReviewStep = ({
       <button
         className="back-button"
         onClick={() => onboardingBag.goTo('benefits')}
+        disabled={onboardingBag.isEmploymentInFinalState}
       >
         Edit Benefits
       </button>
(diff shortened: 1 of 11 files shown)


src/flows/Onboarding/OnboardingBack.tsx
```diff
@@ -3,39 +3,44 @@ import { ButtonHTMLAttributes, PropsWithChildren } from 'react';
...
+    onboardingBag: { back, isEmploymentInFinalState },
   } = useOnboardingContext();
 
   const { components } = useFormFields();
 
+  const onBackHandler = (evt: React.MouseEvent<HTMLButtonElement>) => {
+    if (!isEmploymentInFinalState) {
+      back();
+    } else {
+      onBackError(
```
gabrielseco, line 26: I have the doubts about providing this

This let you know if you haven't put a disabled into the button why the back functionality didn't work

example/src/ReviewStep.tsx
```diff
@@ -199,6 +201,7 @@ export const ReviewStep = ({
       <button
         className="back-button"
         onClick={() => onboardingBag.goTo('basic_information')}
+        disabled={onboardingBag.isEmploymentInFinalState}
```
gabrielseco, line 204: https://github.com/remoteoss/remote-flows/pull/233/files#r2142028922 continuing

For example the goTo doesn't have this functionality, if you don't put the disable, the goTo will work but the effect that we put yesterday, it will move you back to the last step

If you ask me, I am guided on removing the onBackError callback

But I want to know what you think

src/flows/Onboarding/hooks.ts
```diff
@@ -568,5 +566,11 @@ export const useOnboarding = ({
      * Employment data
      */
     employment,
+
+    /**
+     * tells you if an employment in a final state and no more actions can be performed
+     * @returns {boolean}
+     */
+    isEmploymentInFinalState,
```
eng-almeida, line 574: IMO `isEmploymentInFinalState` is not a great name 🤔  maybe isEmploymentReadyForReview? The JSDoc should also be improved. It should describe when this boolean it true

gabrielseco: I went with isEmploymentReadOnly as the employment cannot be edited, open to discuss more in case you don't agree

eng-almeida: Works for me 👍

eng-almeida (approved):

PR #248, opened 2025-06-19
feat(onboarding)- handle errors

gabrielseco:
This in an idea of how the errors could look like.

Right now we only show a generic error, this code is only present in the basic information step as it's a breaking change but probably it should be in all steps and flows

image


src/flows/Onboarding/BasicInformationStep.tsx
```diff
@@ -37,10 +50,24 @@ export function BasicInformationStep({
...
       if (response?.error) {
-        onError?.(response?.error);
+        // Normalize field errors using the utility function
+        const normalizedFieldErrors = normalizeFieldErrors(
+          response?.fieldErrors || [],
+          onboardingBag.meta?.fields?.basic_information,
+        );
+
+        onError?.({
+          error: response?.error,
```
gabrielseco, line 60: why having three properties

error -> is the new Error('Something went wrong')
rawError -> in case people want to do whatever they want with the error returned by the endpoint
fieldErrors -> an object array compose of field and message

One thing that we do is normalize the fieldErrors and the meta tags to assign the correct label

src/lib/mutations.ts
```diff
@@ -13,13 +13,58 @@ export interface SuccessResponse<D> {
...
+              field: nestedKey,
+              messages: nestedValue.map((msg: any) => String(msg)),
+            });
+          }
+        });
+      }
+    });
+  }
+  return fieldErrors;
+}
```
dragidavid, line 66: suggestion:

I think you could make this a little simpler by returning early if errors is not an object. It would also get rid of the `else if`

```suggestion
export function extractFieldErrors(error: any): FieldError[] {
  if (!error || !error.errors || typeof error.errors !== 'object') return [];

  const fieldErrors: FieldError[] = [];

  Object.entries(error.errors).forEach(([key, value]) => {
    if (Array.isArray(value)) {
      fieldErrors.push({
        field: key,
        messages: value.map((msg: any) => String(msg)),
      });
      return;
    }

    if (value && typeof value === 'object') {
      Object.entries(value).forEach(([nestedKey, nestedValue]) => {
        if (Array.isArray(nestedValue)) {
          fieldErrors.push({
            field: nestedKey,
            messages: nestedValue.map((msg: any) => String(msg)),
          });
        }
      });
    }
  });

  return fieldErrors;
}
```

dragidavid (commented):
Having 3 different "types" of errors definitely adds some complexity but the code itself and how you handle them looks good.

Left 1 suggestion but otherwise it seems good!

PR #250, opened 2025-06-23
feat(onboarding) - add functionality to only use employmentId to recover an employment

gabrielseco:
I was asked to add some functionality to recover an employmentId without passing a countryCode.

The countryCode before was passed to initialize the form when we don't have an employmentId and to control which steps we show to the user.

Now countryCode is only to set the country of employment, skipSteps is the prop which makes sure to hide the steps

Besides, we have improved how we call the countries endpoint as we were making always the call, now only when the step appears.


src/flows/Onboarding/tests/OnboardingFlow.test.tsx
```diff
@@ -39,6 +39,13 @@ import userEvent from '@testing-library/user-event';
 
 const queryClient = new QueryClient();
 
+// Helper function to generate unique employment IDs for each test
+let employmentIdCounter = 0;
```
gabrielremote, line 43: I found that it's better to not have a shared mock to make msw to make the call in each test.

That's why we have waitForElementToBeRemoved the spinner to be removed

before the only tests required that we add that, so now it makes sense to me

example/src/OnboardingWithCustomBenefits.tsx
```diff
@@ -208,20 +208,24 @@ type OnboardingWithCustomBenefitsProps = {
...
   type,
   companyId,
   employmentId,
+  countryCode,
 }: OnboardingWithCustomBenefitsProps) => {
   return (
     <RemoteFlows auth={fetchToken}>
       <OnboardingFlow
         companyId={companyId}
+        countryCode={countryCode}
```
gabrielremote, line 224: I detected that I wasn't passing a countryCode here, it's needed

src/flows/Onboarding/api.ts
```diff
@@ -449,8 +450,10 @@ const useCountries = () => {
   });
 };
 
-export const useCountriesSchemaField = (options?: FlowOptions) => {
-  const { data: countries, isLoading } = useCountries();
+export const useCountriesSchemaField = (
+  options?: FlowOptions & { queryOptions?: { enabled?: boolean } },
```
gabrielremote, line 454: this allows to only make the countries call when the country step is loaded

dragidavid (approved):
Happy to approve! Thanks for the comments & explanation!

src/flows/Onboarding/hooks.tsx
```diff
@@ -311,27 +371,20 @@ export const useOnboarding = ({
...
-
-  const isNavigatingToReview = Boolean(
-    employmentId &&
-      isEmploymentReadOnly &&
-      !initialLoading &&
-      stepFields['basic_information'].length > 0 &&
-      stepFields['contract_details'].length > 0 &&
-      stepState.currentStep.name !== 'review',
-  );
+  const { isLoading, isNavigatingToReview, isEmploymentReadOnly } =
```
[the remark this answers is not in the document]
dragidavid: 👌

PR #258, opened 2025-07-01
feat(RemoteFlows): Add support to environment prop

eng-almeida:
Replace the deprecated `isTestingMode` boolean prop with the new `environment` prop, providing users with more granular control over Remote API base url.

### **Benefits:**

**More explicit**: Users can now specify exactly which environment they want to connect to instead of a vague "testing mode"

**Better developer experience**: Clear environment options 

**Flexibility**: Supports all available Remote environments (partners, production, sandbox, staging) rather than just production/sandbox toggle

### **Note:**

With this change `VITE_REMOTE_GATEWAY` env variable by the `example` project needs to be one of the values: production, staging, partners or sandbox.

### DIFF
--- a/example/server.js
+++ b/example/server.js
@@ -5,6 +5,13 @@ import { createServer as createViteServer } from 'vite';
 
 dotenv.config();
 
+const ENVIRONMENTS = {
+  partners: 'https://gateway.partners.remote-sandbox.com',
+  production: 'https://gateway.remote.com',
+  sandbox: 'https://gateway.remote-sandbox.com',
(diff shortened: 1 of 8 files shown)


example/server.js
```diff
@@ -5,6 +5,13 @@ import { createServer as createViteServer } from 'vite';
 
 dotenv.config();
 
+const ENVIRONMENTS = {
```
gabrielseco, line 8: I was thinking why we don't import it with the package but I see in the server env we have an import problem

gabrielseco (approved):

PR #259, opened 2025-07-01
feat: add support for deploying to vercel.

cammellos:
This commit adds support for deploying to vercel.

I have changed the example project to `commonjs`, since that's easier to support on vercel.

I have also removed the dependency on axios, as the package wasn't being imported, but I can revert that change if necessary. I have split index.js and dev_server.js as we don't want to run vite when deploying to vercel.
I have added a flake.nix and a flake.lock, if useful I can keep them, otherwise I am happy to remove them.

App is deployed here https://remote-flows-8d1uny1nr-remotecom.vercel.app/?demo=with-results-cost-calculator , feel free to contact me directly for credentials.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -11,7 +11,7 @@ create an .env file
 
 ```sh
 cd example
-echo "CLIENT_ID=\nCLIENT_SECRET=\nREMOTE_GATEWAY=" >> .env
+echo "CLIENT_ID=\nCLIENT_SECRET=\nVITE_REMOTE_GATEWAY=\nREFRESH_TOKEN=\n" >> .env
 ```
 
 Complete the env file with your env variables and you should be able to run the next step
--- a/example/dev_server.js
+++ b/example/dev_server.js
@@ -1,57 +1,10 @@
-import axios from 'axios';
-import dotenv from 'dotenv';
-import express from 'express';
-import { createServer as createViteServer } from 'vite';
+const axios = require('axios');
+const dotenv = require('dotenv');
+const express = require('express');
+const { getToken } = require('./api/get_token.js');
+const { createServer: createViteServer } = require('vite');
 
 dotenv.config();
-
-const ENVIRONMENTS = {
-  partners: 'https://gateway.partners.remote-sandbox.com',
-  production: 'https://gateway.remote.com',
-  sandbox: 'https://gateway.remote-sandbox.com',
-  staging: 'https://gateway.niceremote.com',
-};
-
-const getToken = async (req, res) => {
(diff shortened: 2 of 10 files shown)


example/dev_server.js
```diff
@@ -1,48 +1,11 @@
-import axios from 'axios';
-import dotenv from 'dotenv';
-import express from 'express';
-import { createServer as createViteServer } from 'vite';
+const axios = require('axios');
```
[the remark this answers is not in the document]
cammellos: oh you are not right, sorry, I removed the dependency only on the code that run on vercel.

example/README.md
```diff
@@ -11,7 +11,7 @@ create an .env file
 
 ```sh
 cd example
-echo "CLIENT_ID=\nCLIENT_SECRET=\nREMOTE_GATEWAY=" >> .env
+echo "CLIENT_ID=\nCLIENT_SECRET=\nVITE_REMOTE_GATEWAY=\nREFRESH_TOKEN=\n" >> .env
```
gabrielseco, line 14: good catch

example/package.json
```diff
@@ -2,23 +2,25 @@
...
     "build": "tsc -b && vite build",
     "lint": "eslint .",
     "preview": "vite preview",
     "postinstall": "react-flagpack"
   },
   "dependencies": {
     "@radix-ui/react-collapsible": "1.1.2",
-    "@remoteoss/remote-flows": "0.4.0-alpha.12",
+    "@remoteoss/remote-flows": "github:remoteoss/remote-flows#main",
     "axios": "^1.8.3",
```
gabrielseco, line 16: You mentioned in the description something about axios getting removed but it's used in the dev_server.js

Or I am missing something?

example/flake.nix
```diff
@@ -0,0 +1,23 @@
+{
+  description = "Remote-flows sdk example project";
```
gabrielseco, line 2: I'm not really familiar with nix

Does it help us to deploy to vercel?

gabrielseco (approved):
I observed a warning from vite about commonjs getting dropped for next version but let's worry about that in the future

example/package.json
cammellos: sorry, i removed it only from `getToken` function, that is the one that is used on vercel, locally it's still used

example/flake.nix
gabrielseco: I don't have hard feelings against this.

Thanks for the clarification

Usually I have node installed globally

gabrielseco:
The merge conflicts are with a PR that I pushed, it should be easy to fix @cammellos

PR #266, opened 2025-07-02
Add currency to the annual gross salary fields

gabrielseco:
When using custom components, the developer will like to know which currency the field is using.

We pass a currency to each field and devs can access to it through fieldData

### DIFF
--- a/src/flows/Onboarding/AnnualGrossSalary.tsx
+++ b/src/flows/Onboarding/AnnualGrossSalary.tsx
@@ -151,19 +151,11 @@ export const AnnualGrossSalary = ({
 
   const { annual_gross_salary_conversion_properties: conversionField } = props;
 
-  const annualGrossSalaryProps = {
-    ...props,
-    ...{ currency: currency },
-  };
-
-  const conversionFieldProps = {
-    ...{ currency: desiredCurrency },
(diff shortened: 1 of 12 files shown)


src/flows/Onboarding/AnnualGrossSalary.tsx
```diff
@@ -151,10 +151,19 @@ export const AnnualGrossSalary = ({
...
+
+  const conversionFieldProps = {
+    ...{ currency: desiredCurrency },
+  };
+
   return (
     <>
       <TextField
-        {...props}
+        {...annualGrossSalaryProps}
```
eng-almeida, line 166: So with this we're passing a `currency` prop to `TextField`, right? IMO this prop is to specific to be passed to a generic component, like TextField. I think would rather have `additionalProps` prop in TextField, that we could use to pass anything relevant to the component. Wdyt?

```
<TextField
   additionalProps={
     {...annualGrossSalaryProps}
   }
  ...
/>
```

gabrielseco: better as it was doing a hack to pass unknown props

eng-almeida (approved):

PR #281, opened 2025-07-04
feat: use options for pdf export

cammellos:
This commit adds the ability to pass options to the cost calculator exports, so that they can benefit of the premium benefits/breakdowns features.


gabrielseco (approved):
Looks good!

PR #300, opened 2025-07-08
feat(cost-calculator) - style premium benefits

gabrielseco:
Style premium benefits adding the currency and style a bit the rest of the components used here.

I went with custom components as it looks like the fastest way to style the currency

Besides that, did a change to the types as there were some types missing

image

### DIFF
--- a/example/src/css/main.css
+++ b/example/src/css/main.css
@@ -579,3 +579,115 @@ button.reset-button {
     0 0 0 3px rgba(5, 150, 105, 0.08),
     0 1px 2px rgba(0, 0, 0, 0.05);
 }
+
+/* ===== INPUT WITH CURRENCY STYLES ===== */
+.input-with-currency {
+  position: relative;
+  display: flex;
+  align-items: stretch;
+  border: 1px solid #e2e8f0;
+  border-radius: 6px;
+  background-color: #ffffff;
+  transition: all 0.2s ease-in-out;
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.05);
+  overflow: hidden;
+}
+
+.input-with-currency:hover {
+  border-color: #cbd5e1;
+  box-shadow: 0 1px 3px rgba(0, 0, 0, 0.08);
+}
+
+.input-with-currency:focus-within {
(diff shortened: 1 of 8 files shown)


src/types/remoteFlows.ts
```diff
@@ -57,7 +57,7 @@ export type FieldComponentProps = {
    * Metadata derived from JSON schema parsing that provides additional context and validation rules for the field.
    * Contains properties defined in the original JSON schema such as type, format, constraints, etc.
    */
-  fieldData: Partial<JSFField>;
+  fieldData: Partial<JSFField> & { currency?: string };
```
eng-almeida, line 60: Why not including currency in `JSFField` type?

example/src/css/main.css
```diff
@@ -579,3 +579,237 @@ button.reset-button {
     0 0 0 3px rgba(5, 150, 105, 0.08),
     0 1px 2px rgba(0, 0, 0, 0.05);
 }
+
+/* ===== INPUT WITH CURRENCY STYLES ===== */
```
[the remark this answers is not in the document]
gabrielseco: I'll create a dev experience project internally to create issues about this

src/types/remoteFlows.ts
eng-almeida: I see! In that case, I think that currency should be part of a `metadata` (or similar) prop or something so these specific props can be placed in there, instead of being in the root of fieldData which is tailored for data coming from json-schemas.

```
Partial<JSFField> & { metadata?: Record<string, unkwnow> };
```

eng-almeida: Oh right! I didn't realise about that change in the previous PR, sorry! Well this is not a blocker, so if you feel the best for now is to proceed without a breaking change, then feel free to ignore this for now and we can address it afterwards.

gabrielseco: I feel they're not there yet as I experience type errors in the consumer app when accessing to the currency property

The change will be easy in my opinion is changing one line

eng-almeida (approved):

PR #314, opened 2025-07-11
fix(fieldsets): Trigger manual validation

eng-almeida:
This PR fixes a bug where the error state is not properly updated when a field is inside a fieldset. Even though these fields have no errors, the UI would continue to indicate that an error existed.

The root cause of this behavior with fieldsets isn't entirely clear, so the solution involves triggering manual validation for fieldsets whenever one of their values changes.

### DIFF
--- a/src/components/form/fields/FieldSetField.tsx
+++ b/src/components/form/fields/FieldSetField.tsx
@@ -3,6 +3,8 @@ import { cn } from '@/src/lib/utils';
 import { SupportedTypes } from './types';
 import { Components } from '@/src/types/remoteFlows';
 import { Statement, StatementProps } from '@/src/components/form/Statement';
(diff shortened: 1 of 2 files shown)


src/components/form/fields/FieldSetField.tsx
```diff
@@ -39,6 +41,51 @@ export function FieldSetField({
...
+      if (triggerTimeoutRef.current) {
+        clearTimeout(triggerTimeoutRef.current);
+      }
+      triggerTimeoutRef.current = setTimeout(() => {
+        trigger();
+      }, 50);
+
+      return;
+    }
+    prevValuesRef.current = [...currentValues];
```
cammellos, line 80: this would not be hit if `hasChanged` is true, is that intentional?

cammellos (approved):

PR #322, opened 2025-07-17
fix(onboarding): enable ack when an employment id exists

eng-almeida:
Some ack fields are not stored in the database, such as "ack_start_date_ammendment" for Argentina, and therefore are not returned in the employment response. Ideally, the solution should be implemented on the backend, with the employment response including these fields. For now, the solution is to enable the ack fields whenever an employmentId exists (for example, when an onboarding is saved as a draft), as this indicates that the user has already started the onboarding process.

### DIFF
--- a/src/components/form/utils.ts
+++ b/src/components/form/utils.ts
@@ -760,7 +760,7 @@ export function enableAckFields(
 ) {
   let result = values;
   fields.forEach((field) => {
-    if (typeof field === 'object' && field !== null && 'const' in field) {
+    if ('const' in field) {
       result = Object.fromEntries(
         Object.entries(values).map(([k, v]) => {
           if (k === field.name) {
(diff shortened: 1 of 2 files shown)


src/components/form/utils.ts
```diff
@@ -754,6 +754,28 @@ export function getFieldsWithFlatFieldsets({
   return filteredFields;
 }
 
+export function enableAckFields(
+  fields: Fields,
+  values: Record<string, unknown>,
+) {
+  let result = values;
```
gabrielseco, line 761: I have a different way but I don't know if I am missing something

```ts
let result: Record<string, unknown> = {};
  fields.forEach((field) => {
    if ('const' in field) {
      result = Object.fromEntries(
        Object.entries(values).map(([key, value]) => {
          if (key === field.name) {
            return [key, field.const];
          }
          return [key, value];
        }),
      );
    }
  });
  return result;
```

Why do you assign values directly to the result?
Why the Object.keys, cannot be done, with fields.forEach?
Does my conditional 'const' in field already checks for objects and nulls right?

PR #350, opened 2025-07-25
feat(cost-calculator) - reset form fields explicitly

gabrielseco:
I did a new feature in the CostCalculator to only reset certain fields

In this case, we want to reset only country but not salary or currency

### DIFF
--- a/src/flows/CostCalculator/CostCalculatorForm.tsx
+++ b/src/flows/CostCalculator/CostCalculatorForm.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from 'react';
 import { CostCalculatorEstimateResponse } from '@/src/client';
 import { JSONSchemaFormFields } from '@/src/components/form/JSONSchemaForm';
 import { Form } from '@/src/components/ui/form';
@@ -34,7 +35,7 @@ type CostCalculatorFormProps = Partial<{
   /**
    * Fields to reset when the form is successfully submitted.
    */
-  shouldResetFormFields?: string[];
+  resetFields?: ('country' | 'currency' | 'salary')[];
 }>;
 
 export function CostCalculatorForm({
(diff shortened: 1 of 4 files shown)


src/flows/CostCalculator/CostCalculatorForm.tsx
```diff
@@ -54,9 +60,23 @@ export function CostCalculatorForm({
...
+        // Reset only the form fields
+        if (shouldResetFormFields) {
+          const currentValues = form.getValues();
+          const resetValues = { ...currentValues };
+          shouldResetFormFields.forEach((field) => {
+            resetValues[field] = '';
+          });
+
+          costCalculatorBag?.resetForm();
+          form.reset(resetValues);
```
eng-almeida, line 79: According to the [docs](https://react-hook-form.com/docs/useform/reset), check `Submit with Reset` section, the reset should be done inside of a useEffect.

src/flows/CostCalculator/CostCalculatorForm.tsx
```diff
@@ -30,13 +30,19 @@ type CostCalculatorFormProps = Partial<{
    * Whether to reset the form when the form is successfully submitted.
    */
   shouldResetForm?: boolean;
+
+  /**
+   * Fields to reset when the form is successfully submitted.
+   */
+  shouldResetFormFields?: string[];
```
eng-almeida, line 37: Can't we change the type to an array of enum's?

src/flows/CostCalculator/CostCalculatorForm.tsx
```diff
@@ -30,13 +30,19 @@ type CostCalculatorFormProps = Partial<{
...
 }>;
 
 export function CostCalculatorForm({
   onSubmit,
   onError,
   onSuccess,
   shouldResetForm,
+  shouldResetFormFields,
```
eng-almeida, line 45: We should find a better name, because it sound like it's a boolean, when in fact it's a list. Maybe `resetFields`?

src/flows/CostCalculator/CostCalculatorForm.tsx
```diff
@@ -54,9 +60,23 @@ export function CostCalculatorForm({
     } else {
       if (costCalculatorResults?.data) {
         await onSuccess?.(costCalculatorResults?.data);
+        // Reset the form completely
         if (shouldResetForm) {
```
eng-almeida, line 64: if `shouldResetForm` is true and `shouldResetFormFields` is defined, then `shouldResetForm` takes precedence. Is this the expected behaviour? If so, we should update the README.

src/flows/CostCalculator/README.md
```diff
@@ -555,12 +555,17 @@ The options.jsfModify props accepts the same props that the [modify](https://jso
...
+| `onSubmit`        | `(payload: CostCalculatorEstimateParams) => void`    | No         | Callback with the form payload sent to Remote API. Runs before submitting the form to Remote |
+| `onSuccess`       | `(response: CostCalculatorEstimateResponse) => void` | No         | Callback with the successful estimation data                                                 |
+| `onError`         | `(error: Error) => void`                             | No         | Error handling callback                                                                      |
+| `shouldResetForm` | `boolean`                                            | No         | If true, the form will be reset after a successful submission. Default is `false`.           |
+| `resetFields`     | `'country'                                           | 'currency' | 'salary'[]`                                                                                  | No  | You can pass an array like ['country'] and only country and the region will be cleared |
```
eng-almeida:
```suggestion
| `resetFields`     | `'country'                                           | 'currency' | 'salary'[]`                                                                                  | No  | Clears specified form fields. Pass field names as an array (e.g., ['country']). When resetting nested fields like 'country', related dependent fields (e.g., region) will also be cleared automatically. |
```

PR #370, opened 2025-08-04
zendesk articles

gabrielseco:


example/src/css/main.css
```diff
@@ -877,3 +877,25 @@ button.reset-button {
   font-size: 14px;
   letter-spacing: 0.01em;
 }
+
+.RemoteFlows_ZendeskDrawer__Content {
```
gabrielseco, line 881: We put some styles to the content of the drawer that is inserted dynamically with the API

src/client/types.gen.ts
```diff
@@ -86,15 +86,21 @@ export type MinimalRegion = {
 export type CountryCode = string;
 
 export type TerminationOffboarding = {
```
gabrielseco, line 88: This file is autogenerated with the openapi:ts, it just generates types based on our staging open api spec

src/components/shared/zendesk-drawer/api.ts
```diff
@@ -0,0 +1,34 @@
+import { getShowHelpCenterArticle } from '@/src/client';
+import { useClient } from '@/src/context';
+import { Client } from '@hey-api/client-fetch';
+import { useQuery } from '@tanstack/react-query';
+
+export const useZendeskArticle = (
```
gabrielseco, line 6: as the component is the only one that calls this endpoint, I put the fetcher here

src/components/shared/zendesk-drawer/ZendeskDrawer.tsx
```diff
@@ -0,0 +1,78 @@
...
+  Trigger,
+  zendeskId,
+  zendeskURL,
+}: ZendeskDrawerProps) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const searchParams = useSearchParams();
+  const router = useRouter();
+
+  useEffect(() => {
+    const articleId = searchParams.get('articleId');
```
gabrielseco, line 31: Once the articleId is in the URL, we made the call to the API to load the content and open the drawer

src/components/ui/form.tsx
```diff
@@ -13,7 +12,7 @@ import {
 } from 'react-hook-form';
 
 import { Label } from '@/src/components/ui/label';
-import { cn } from '@/src/lib/utils';
+import { cn, sanitizeHtml } from '@/src/lib/utils';
```
gabrielseco, line 15: we move the code to a util to be able to reuse it in the drawer

src/flows/CostCalculator/Results/CostCalculatorResults.tsx
```diff
@@ -169,7 +172,8 @@ export function CostCalculatorResults({
                 currency={currency}
               />
               {/* Extra Statutory Payments Section */}
-              {extraStatutoryPaymentsBreakdown ? (
+              {Array.isArray(extraStatutoryPaymentsBreakdown) &&
```
gabrielseco, line 175: fix a bug that we had in the extra statuory payments, we need to check if the data is an array and bigger than 0

src/lib/router.ts
```diff
@@ -0,0 +1,46 @@
+import { useCallback } from 'react';
```
gabrielseco, line 1: I created router.ts and useSearchParams to abstract the code related to the window.history.

The SDK doesn't have a router like nextjs and router-react-dom as we need to be agnostic.

the API I followed is similar to NextJs so everyone is familiar with it

src/components/shared/zendesk-drawer/api.ts
```diff
@@ -0,0 +1,34 @@
...
+  queryOptions?: { enabled?: boolean },
+) => {
+  const { client } = useClient();
+  return useQuery({
+    queryKey: ['zendesk-article', zendeskId],
+    queryFn: async () => {
+      const response = await getShowHelpCenterArticle({
+        client: client as Client,
+        headers: {
+          Authorization: ``,
```
remotecom, line 17: why is this Auhtorization empty ?

PR #377, opened 2025-08-07
fix(force-value) - fix undefined statement?.title

gabrielseco:
We had a report that for belgium contract details the title attribute was undefined.

This is due to the json-schema that it doesn't define inside the statement and we need to use the property title from json-schema which json-schema-form maps it to label

image


src/components/form/fields/tests/ForcedValueField.test.tsx
```diff
@@ -0,0 +1,220 @@
+import { render, screen } from '@testing-library/react';
```
gabrielseco, line 1: added tests to avoid regressions in the future

src/components/form/fields/ForcedValueField.tsx
```diff
@@ -6,16 +7,18 @@ type ForcedValueFieldProps = {
   value: string;
   description: string;
   statement?: {
-    title: string;
+    title?: string;
```
gabrielseco, line 10: The title can be undefined, that's why we change the type

PR #389, opened 2025-08-13
feat(premium-benefits) - add toolbar

gabrielseco:
I reorder the code in different views to be able to transitionate between the different views

I added the toolbar outside of the library, once I get the clear idea how BHR wants to embed that, I'll propose a solution.

The only part that is new in the library is the exported EstimationError which wasn't done before

image

### DIFF
--- a/src/flows/CostCalculator/CostCalculatorForm.tsx
+++ b/src/flows/CostCalculator/CostCalculatorForm.tsx
@@ -3,10 +3,10 @@ import { CostCalculatorEstimateResponse } from '@/src/client';
 import { JSONSchemaFormFields } from '@/src/components/form/JSONSchemaForm';
 import { Form } from '@/src/components/ui/form';
 import { useCostCalculatorContext } from '@/src/flows/CostCalculator/context';
-import { EstimationError } from '@/src/flows/CostCalculator/hooks';
 import {
   CostCalculatorEstimationFormValues,
   CostCalculatorEstimationSubmitValues,
+  EstimationError,
 } from '@/src/flows/CostCalculator/types';
 
 type CostCalculatorFormProps = Partial<{


example/src/CostCalculatorWithPremiumBenefits.tsx
```diff
@@ -26,6 +27,166 @@ const estimationOptions = {
...
+                description:
+                  'Select the country where the employee will primarily live and work',
+              },
+              currency: {
+                title: 'Employer billing currency',
+                description:
+                  'Select the currency you want to be invoiced in for this employee’s services.',
+              },
+              salary: {
+                title: "Employee's annual salary",
```
cammellos, line 121: very minor neatpick: there's some inconsistency on quotes usage, you use forward tick below, while use ' here

example/src/CostCalculatorWithPremiumBenefits.tsx
```diff
@@ -26,6 +27,166 @@ const estimationOptions = {
...
+    <>
+      <Header />
+      <div className="mb-8">
+        <ActionToolbar
+          onReset={onReset}
+          onExportPdf={onExportPdf}
+          onCSVExport={() => {}}
+        />
+      </div>
+      <div className="mt-4 mb-2 flex gap-2">
```
cammellos, line 179: If there's no employment, should we show this?
i.e

```
const primaryEmployment = estimations.data.employments?.[0];
const country = primaryEmployment?.country;

{country && (
  <div className="mt-4 mb-2 flex gap-2">
    <Flag code={country.alpha_2_code} />
    <span className="text-md font-bold">{country.name}</span>
  </div>
)}
```

cammellos (approved):
overall looks ok, left two minor comments

PR #405, opened 2025-08-20
feat(cost-calculator) - call new conversion endpoints

gabrielseco:
We want that our platform and the SDK will be more similar when doing conversions.

The raw endpoint supports a similar conversion and it's supposed to work for the Onboarding and Marketing Cost Calculator

The effective endpoint is a replacement of the previous endpoint used

### DIFF
--- a/src/flows/Onboarding/api.ts
+++ b/src/flows/Onboarding/api.ts
@@ -502,23 +502,14 @@ export const useConvertCurrency = ({
   const { client } = useClient();
   return useMutation({
     mutationFn: (payload: ConvertCurrencyParams) => {
-      if (type === 'no_spread') {
-        return postConvertRawCurrencyConverter({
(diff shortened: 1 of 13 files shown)


src/flows/Onboarding/api.ts
```diff
@@ -493,15 +494,20 @@ export const useCountriesSchemaField = (
...
   return useMutation({
     mutationFn: (payload: ConvertCurrencyParams) => {
-      return postConvertCurrencyConverter({
+      const apiFn =
+        type === 'no_spread'
+          ? postConvertRawCurrencyConverter
+          : postConvertWithSpreadCurrencyConverter;
+      return apiFn({
         client: client as Client,
-        headers: {
```
eng-almeida, line 502: The header is missing 🙂

PR #408, opened 2025-08-20
feat: use annual_gross_salary_in_employer_currency

cammellos:
When the user types the salary into the employer currency, we should use that one instead of the converted amount when estimating on the backend, so that the results are consistent.

To do that:
- We keep track of whether the user has typed into the conversion input
- We surface that information to the submit value, and we set either annual_gross_salary or annual_gross_salary_in_employer_currency based on whether the user has inputted in one or the other currency.

### DIFF
--- a/src/components/form/fields/CurrencyConversionField.tsx
+++ b/src/components/form/fields/CurrencyConversionField.tsx
@@ -163,8 +163,12 @@ export const CurrencyConversionField = ({
       ),
     500,
   );
+  // we keep track of the last input the user used, so we can make sure
+  // we keep consistent currency rates
+  const lastInputFieldName = `${props.name}_converted`;
 
   const handleMainFieldChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
+    setValue(lastInputFieldName, false);
     if (showConversion) {
       debouncedConvertCurrency(evt.target.value);
     }
@@ -173,6 +177,7 @@ export const CurrencyConversionField = ({
   const handleConversionFieldChange = (
     evt: React.ChangeEvent<HTMLInputElement>,
   ) => {
+    setValue(lastInputFieldName, true);
     debouncedConvertCurrencyReverse(evt.target.value);
   };
 
@@ -222,6 +227,11 @@ export const CurrencyConversionField = ({
           onChange={handleConversionFieldChange}
         />
       )}
+      <input
+        type="hidden"
+        name={lastInputFieldName}
+        value={watch(lastInputFieldName) || false}
+      />
     </>
   );
 };
--- a/src/flows/CostCalculator/CostCalculatorForm.tsx
+++ b/src/flows/CostCalculator/CostCalculatorForm.tsx
@@ -6,8 +6,10 @@ import { useCostCalculatorContext } from '@/src/flows/CostCalculator/context';
 import {
   CostCalculatorEstimationFormValues,
   CostCalculatorEstimationSubmitValues,
+  CostCalculatorInternalEstimationSubmitValues,
   EstimationError,
 } from '@/src/flows/CostCalculator/types';
+import omit from 'lodash/omit';
 
 type CostCalculatorFormProps = Partial<{
   /**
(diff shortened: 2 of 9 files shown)


src/flows/CostCalculator/utils.ts
```diff
@@ -53,14 +53,17 @@ function mapValueToEmployment(
...
   return {
     region_slug: value.region || value.country,
-    annual_gross_salary_in_employer_currency: value.salary,
     employment_term: value.contract_duration_type ?? 'fixed',
     title: estimationOptions.title,
     age: value.age ?? undefined,
     ...(value.benefits && { benefits: formatBenefits(value.benefits) }),
-    ...(version === 'standard' && {
-      annual_gross_salary: value.salary,
+    ...(value.salary_converted && {
```
gabrielseco, line 60: unless I am mistaken we need to send it when value.salary_converted or version === 'marketing'

If you have set up the version prop marketing and salary_converted is false, we'll never send this prop

src/components/form/fields/CurrencyConversionField.tsx
```diff
@@ -218,6 +223,11 @@ export const CurrencyConversionField = ({
           onChange={handleConversionFieldChange}
         />
       )}
+      <input
+        type="hidden"
+        name={lastInputFieldName}
+        value={watch(lastInputFieldName) || false}
```
gabrielseco, line 229: Nitpick: instead of boolean I would send strings

for example the default value can be props.name

src/components/form/fields/CurrencyConversionField.tsx
```diff
@@ -169,6 +173,7 @@ export const CurrencyConversionField = ({
   const handleConversionFieldChange = (
     evt: React.ChangeEvent<HTMLInputElement>,
   ) => {
+    setValue(lastInputFieldName, true);
```
gabrielseco, line 176: and here as arg conversionFieldName

then in the rest of the code you could identify the strings that we send when parse values and it would be easy to read

src/components/form/fields/CurrencyConversionField.tsx
cammellos: would it be ok if i leave it as a boolean for now? i could address in a different pr if that's ok

remotecom: I don't mind leaving the boolean for now

gabrielseco: I don't mind

src/flows/CostCalculator/utils.ts
```diff
@@ -53,14 +53,17 @@ function mapValueToEmployment(
...
   return {
     region_slug: value.region || value.country,
-    annual_gross_salary_in_employer_currency: value.salary,
     employment_term: value.contract_duration_type ?? 'fixed',
     title: estimationOptions.title,
     age: value.age ?? undefined,
     ...(value.benefits && { benefits: formatBenefits(value.benefits) }),
-    ...(version === 'standard' && {
-      annual_gross_salary: value.salary,
+    ...((version == 'marketing' || value.salary_converted) && {
```
gabrielseco, line 60: we need to add the salary_converted type in CostCalculatorEstimationSubmitValues I believe

src/flows/CostCalculator/CostCalculatorForm.tsx
```diff
@@ -80,13 +80,13 @@ export function CostCalculatorForm({
...
-    const cleanedValues = costCalculatorBag?.parseFormValues(
+    const parsedValues = costCalculatorBag?.parseFormValues(
       values,
     ) as CostCalculatorEstimationSubmitValues;
     const costCalculatorResults =
-      await costCalculatorBag?.onSubmit(cleanedValues);
+      await costCalculatorBag?.onSubmit(parsedValues);
 
-    await onSubmit?.(cleanedValues);
+    await onSubmit?.(parsedValues);
```
gabrielseco, line 89: I think we should clean the parsedValues, I don't think there's a benefit for the partner to know if salary_converted is true or false

That way we don't have to change the CostCalculatorFlow tests

cammellos: in that case though, I have to `Omit<>` and it gets a bit icky because the type is exported, so we need to change all the types in the example app as well to use omit, or provide a custom type that omits it (example app doesn't compile if we add Omit), (not 100% sure, but i applied your changes, and example app was complaining)

gabrielseco: oh I see, let me have a look

PR #417, opened 2025-08-21
docs(summary) - improve summary results logic && docs

gabrielseco:
Fix the summary results as we weren't grouping by country the estimations.

Add docs && add tests for this one

### DIFF
--- a/src/flows/CostCalculator/SummaryResults/SummaryResults.tsx
+++ b/src/flows/CostCalculator/SummaryResults/SummaryResults.tsx
@@ -12,19 +12,40 @@ import { useState } from 'react';
 
 const useSummaryResults = (estimations: CostCalculatorEmployment[]) => {
   const currency = estimations[0]?.employer_currency_costs.currency;
-  const costsPerCountry = estimations.map((estimation) => {
-    return {
-      country: estimation.country,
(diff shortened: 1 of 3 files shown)


src/flows/CostCalculator/SummaryResults/SummaryResults.tsx
```diff
@@ -11,20 +11,48 @@ import { ChevronDown, Globe } from 'lucide-react';
...
-        estimation.employer_currency_costs.annual_total,
-        currency.symbol,
-      ),
+      currency: null,
+      costsPerCountry: [],
+      employeesCost: null,
     };
-  });
+  }
+  const currency = estimations[0]?.employer_currency_costs.currency;
```
cammellos, line 21: looks like you are allowing currency to be null, but it would fail eventually, should we not just guard against it and return early?

gabrielseco: we're returning null if the currency is not set

https://github.com/remoteoss/remote-flows/pull/417/files#diff-42b32328b5c19f48bd871a6f9d7926e51b4a1a11849c2663b92c14a516bbe235R261

PR #423, opened 2025-08-25
feat(cost-calculator) - add management fieldset

gabrielseco:
Adds management field to the CostCalculator which allows editing the management field in the deployed app.

image

Review app: https://remote-flows-5933259wo-remotecom.vercel.app/

Review the fieldsets change first https://github.com/remoteoss/remote-flows/pull/425


src/flows/CostCalculator/CostCalculatorFlow.tsx
```diff
@@ -44,6 +46,17 @@ export type CostCalculatorFlowProps = {
...
 
+const getDefaultManagementFee = (
+  baseRates: Record<CurrencyKey, number>,
+  currency: CurrencyKey,
+  managementFees?: Record<CurrencyKey, number>,
+) => {
+  if (managementFees && managementFees[currency]) {
+    return managementFees[currency];
+  }
+  return baseRates[currency] / 100;
```
cammellos, line 57: looks like if baseRates[currency] is undefined (i.e we don't provide a default), you return `NaN`, which is falsey, but looks like you might want to be explicit and guard against it, instead of letting the division do the work

PR #436, opened 2025-09-01
grid demo for cost calculator

gabrielseco:


example/src/CostCalculatorWithPremiumBenefits.tsx
```diff
@@ -237,6 +238,9 @@ const AddEstimateForm = ({
               description:
                 "We will use your selected billing currency, but you can also convert it to the employee's local currency.",
             },
+            salary_conversion: {
```
gabrielseco, line 241: this is how you can override the label for the salary_conversion and it always be in the converted field quantity even if we change the fields names

PR #439, opened 2025-09-04
feat: add hiring budget option

cammellos:
This PR adds an option to specify whether you want to use your budget or the employee's annual salary (default).

It also surfaces the error from the API in case an error is returned.

A couple of things:
- I have tried to use JSON schema to change the title of the field, but that didn't work, the same JSON schema works in the playground. To accommodate for that, I am setting it when the value is selected.
- I have added an onChange callback so that I can clear the error whenever any form field changes.
- I have surfaced the error in the UI if the request failed, I could not get the type right though unfortunately.

### DIFF
--- a/example/e2e/hiring-budget.spec.ts
+++ b/example/e2e/hiring-budget.spec.ts
@@ -1,0 +1,26 @@
+import { test, expect } from '@playwright/test';
+import { fillEstimationForm } from './helpers';
+
+test.describe('hiring budget', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/?demo=with-premium-benefits-cost-calculator');
+  });
+
+  test('shows the hiring budget option', async ({ page }) => {
+    await page.locator('#my_hiring_budget').click();
+
+    await expect(page.locator('[for=salary]')).toHaveText('Hiring budget');
+
+    await fillEstimationForm(page, {
+      country: 'Sweden',
+      currency: 'USD',
+      salary: '100000',
+    });
+
+    const headerAmount = page.getByText(
+      /Employee annual gross salary: kr\d{1,3}(,\d{3})*\.\d{2}/,
+    );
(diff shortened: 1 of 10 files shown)


example/e2e/hiring-budget.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    await fillEstimationForm(page, {
+      country: 'Sweden',
+      currency: 'USD',
+      salary: '100000',
+    });
+
+    await expect(
+      page.locator(
+        '[data-selector=estimation-results-header-annual-gross-salary]',
```
gabrielseco, line 22: can you sync with main, I changed data-selector with data-testid

You can use 

```
page.getByTestId(
      'estimation-results-header-annual-gross-salary',
    );
```

Which I think it has better readability

gabrielseco (commented):
A comment for now, testing the PR

example/e2e/hiring-budget.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    await expect(page.locator('[for=salary]')).toHaveText('Hiring budget');
+
+    await fillEstimationForm(page, {
+      country: 'Sweden',
+      currency: 'USD',
+      salary: '100000',
+    });
+
+    const employerAmount = page.getByTestId(
```
gabrielseco, line 20: we need to assert the employerAmount right?

cammellos:
> I see when the hiring budget is chosen and you put 50k, we're losing some cents
> 
> Just an observation but not hard requirement for me to not approve
> image

that's correct, basically it needs to stay within the budget, it will not be able to hit it exactly right

example/e2e/hiring-budget.spec.ts
gabrielseco: ready to ship

PR #441, opened 2025-09-09
feat: show annual gross salary in header

cammellos:
This commit adds the annual gross salary in the header, using the employee currency.

### DIFF
--- a/example/e2e/annual-gross-salary.spec.ts
+++ b/example/e2e/annual-gross-salary.spec.ts
@@ -5,13 +5,20 @@ test.describe('annual gross salary', () => {
     await page.goto('/?demo=with-premium-benefits-cost-calculator');
   });
 
-  test('shows the annual gross salary', async ({ page }) => {
-    await page.selectOption('#country', 'Argentina');
+  test('shows the annual gross salary both in the header and in the table', async ({
+    page,
+  }) => {
+    await page.selectOption('#country', 'Sweden');
     await page.selectOption('#currency', 'USD');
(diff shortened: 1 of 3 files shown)


example/e2e/annual-gross-salary.spec.ts
```diff
@@ -5,13 +5,20 @@ test.describe('annual gross salary', () => {
...
     await page.click('.submit-button');
     await expect(
       page.locator('[data-selector=annual-gross-salary-employer-amount]'),
-    ).toHaveText('$1,000.00');
+    ).toHaveText('$100.00');
+    await expect(
+      page.locator(
+        '[data-selector=estimation-results-header-annual-gross-salary]',
+      ),
+    ).toHaveText(/Employee annual gross salary: kr\d+\.\d\d/);
```
eng-almeida, line 22: I'm assuming we are doing this because we don't know the actual conversion so we're just expecting for a number to be rendered, right? Should we consider mocking API requests or is there a specific reason for using the real API?

eng-almeida: > we would spot the issue immediately

Depends on what you consider immediately 😅 If the backend does a breaking change of that nature you can still deploy it and have no visibility over the SDK. You'll only get visibility on the following PR for the SDK. But I get your point for sure 💯 We should probably discuss internally contract testing.

cammellos: makes sense, i can ping gabril once he gets back, thanks!

PR #442, opened 2025-09-10
bug: hide "Days" if minimum onboarding time is missing

cammellos:
Sometimes the information is not avaiable, so we hide "Days" (we could also at some point make it singular/plural depending on the length, but i would thave to check if there's any mot that is 1)

### DIFF
--- a/src/flows/CostCalculator/EstimationResults/EstimationResults.tsx
+++ b/src/flows/CostCalculator/EstimationResults/EstimationResults.tsx
@@ -154,9 +154,11 @@ function OnboardingTimeline({
             <span className='RemoteFlows__OnboardingTimeline__Title text-base font-medium text-[#0F172A]'>
               Onboarding timeline
             </span>
-            <span className='RemoteFlows__OnboardingTimeline__Description text-base text-muted-foreground mr-4'>
-              {minimumOnboardingDays} days
-            </span>
+            {minimumOnboardingDays != null && (
+              <span className='RemoteFlows__OnboardingTimeline__Description text-base text-muted-foreground mr-4'>
+                {minimumOnboardingDays} days
+              </span>
+            )}
           </div>
         </AccordionTrigger>
         <AccordionContent className='px-0 pb-4'>


src/flows/CostCalculator/EstimationResults/EstimationResults.tsx
```diff
@@ -154,7 +154,7 @@ function OnboardingTimeline({
               Onboarding timeline
             </span>
             <span className='RemoteFlows__OnboardingTimeline__Description text-base text-muted-foreground mr-4'>
-              {minimumOnboardingDays} days
+              {minimumOnboardingDays && `${minimumOnboardingDays} days`}
```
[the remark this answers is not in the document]
cammellos: probably safer to compare with null, i don't think it will be 0, but you never know :), i'll push a fix, thanks

src/flows/CostCalculator/EstimationResults/EstimationResults.tsx
```diff
@@ -154,7 +154,7 @@ function OnboardingTimeline({
               Onboarding timeline
             </span>
             <span className='RemoteFlows__OnboardingTimeline__Description text-base text-muted-foreground mr-4'>
-              {minimumOnboardingDays} days
+              {minimumOnboardingDays != null && `${minimumOnboardingDays} days`}
```
gabrielseco, line 157: can we hide the entire span instead of just the text like i did 

https://github.com/remoteoss/remote-flows/pull/444

src/flows/CostCalculator/EstimationResults/EstimationResults.tsx
```diff
@@ -154,9 +154,11 @@ function OnboardingTimeline({
             <span className='RemoteFlows__OnboardingTimeline__Title text-base font-medium text-[#0F172A]'>
               Onboarding timeline
             </span>
-            <span className='RemoteFlows__OnboardingTimeline__Description text-base text-muted-foreground mr-4'>
-              {minimumOnboardingDays} days
-            </span>
+            {minimumOnboardingDays != null && (
```
gabrielseco, line 157: you can cover undefined && null at the same time by doing

`minimumOnboardingDays && `

PR #444, opened 2025-09-16
fix minimumOnboardingTime

gabrielseco:
For countries like barbados we don't provide a minimumOnboardingTime.

This avoids showing the content if there isn't one


eng-almeida:
There's already a PR open addressing this issue https://github.com/remoteoss/remote-flows/pull/442/files

PR #446, opened 2025-09-16
feat(cost-calculator) - do edit functionality but partially

gabrielseco:
We have some missing pieces of the edit functionality such as 

* not render the employer currency
* render an textfield to change the estimation title

I tried to the edit functionality to what we have now and we can do separated PRs addressing the missing functionality

In the [deployed review app](https://remote-flows-219159v49-remotecom.vercel.app/), you can check the functionality


cammellos (commented):
@gabrielremote could you add an E2E test please validating the behavior?

cammellos (approved):

PR #459, opened 2025-09-19
feat(cost-calculator) - hides currency

gabrielseco:
Hides currency for adding estimate / edit estimate

image

There's still some work do like adding a force field value statement but now works better

### DIFF
--- a/example/src/CostCalculatorWithPremiumBenefits.tsx
+++ b/example/src/CostCalculatorWithPremiumBenefits.tsx
@@ -115,6 +115,7 @@ const DrawerEstimationForm = ({
   Trigger?: React.ReactElement;
   options: {
     title: string;
+    hideCurrency?: boolean;
   };
   defaultValues?: CostCalculatorFlowProps['defaultValues'];
   header: {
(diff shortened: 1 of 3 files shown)


example/src/CostCalculatorWithPremiumBenefits.tsx
```diff
@@ -192,6 +197,7 @@ const EditEstimationForm = ({
         currencySlug: payload?.currency,
         salary: convertFromCents(payload?.salary)?.toString() ?? '',
         benefits: payload?.benefits,
+        selectedCurrency: estimation?.employer_currency_costs.currency.code,
```
cammellos, line 200: Maybe it'd be good to have a var that explains what it is, i.e
```
const previouslySelectedEstimate = estimation?.... 
```
(also, i think there's a bit of value in being a more explicit, `estimation?.employe...` seems to be defensive coding, but here we use it for actually driving the logic, so 
```
const ... = estimation ? estimation.... : null;
```

looks more intentional and explicit

cammellos (approved):

example/src/CostCalculatorWithPremiumBenefits.tsx
gabrielseco: I changed the prop but we can still do estimation? and use ?? operator

PR #465, opened 2025-09-24
add management field test

gabrielseco:
Test to prevent the disappearance of the management field in the deployed app

### DIFF
--- a/example/e2e/add-estimation.spec.ts
+++ b/example/e2e/add-estimation.spec.ts
@@ -6,17 +6,12 @@ test.describe('add estimation from drawer', () => {
     await page.goto('/?demo=with-premium-benefits-cost-calculator');
   });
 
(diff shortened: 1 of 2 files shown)


cammellos (approved):

example/e2e/add-estimation.spec.ts
```diff
@@ -6,6 +6,12 @@ test.describe('add estimation from drawer', () => {
     await page.goto('/?demo=with-premium-benefits-cost-calculator');
   });
 
+  test('should have a management fee field', async ({ page }) => {
```
cammellos, line 9: it's a bit heavy handed for an e2e test, i would maybe include it in a different test or test it at a lower level, but it doesn't hurt of course

gabrielseco: let me rethink it to extract more value

PR #468, opened 2025-09-24
feat: carry over hiring budget when editing

cammellos:


gabrielseco:
Merging it to include it on the release

PR #486, opened 2025-10-02
feat(cost-calculator) - add csv download

gabrielseco:
Add csv download from the API

### DIFF
--- a/example/src/CostCalculatorWithPremiumBenefits.tsx
+++ b/example/src/CostCalculatorWithPremiumBenefits.tsx
@@ -614,7 +614,7 @@ function CostCalculatorFormDemo() {
       {
         onSuccess: (response) => {
           downloadFile(
-            response.data.data.content as unknown as string,
+            response?.data?.data?.content as unknown as string,
             'estimation.csv',
           );
         },


example/src/CostCalculatorWithExportPdf.tsx
```diff
@@ -34,7 +34,7 @@ function CostCalculatorFormDemo() {
       exportPdfMutation.mutate(buildCostCalculatorEstimationPayload(payload), {
         onSuccess: (response) => {
           if (response?.data?.data?.content !== undefined) {
-            downloadPdf(
+            downloadFile(
               response.data.data.content as unknown as string,
               'estimation.pdf',
```
hamzaremote, line 39: the filename will always be .pdf even when it is a csv?

hamzaremote: ah my bad, please ignore my comment

hamzaremote (approved):

example/src/CostCalculatorWithExportPdf.tsx
gabrielseco: I changed the name of the method as we could reuse it for csvs

PR #493, opened 2025-10-07
fix(cost-calculator) - salary conversions when we swap currencies

gabrielseco:
When changing the employer billing currency or even country the salary fields weren't transforming correctly every time.


gabrielseco:
not yet ready, I though it was fixed but it isn't

cammellos (approved):

PR #1001, opened 2026-05-06
fix(money) - parse values before prettifying

gabrielseco:
## The problem
When users use the onboarding for Belgium, the contract details has some fields like eco_voucher and yearly premium rendered incorrectly

image

## The solution
The solution is to pass to prettifyFormValues in the onSubmit, the values already parsed and that way in prettifyFormValues we can convert always trusting that the input will be always in cents

the problem was the values of the form was mixing cents && normal values in different inputs

We builld a escape hatch when we call prettifyFormValues when an employee has been invited

https://www.loom.com/share/2efa29079483435f8c160faef9a7ba81

> [!NOTE]
> **Medium Risk**
> Touches shared form value transformation and review metadata rendering, which can affect how money amounts and some field values are displayed across multiple flows. Main risk is incorrect unit conversion (cents vs. major units) in review screens or tests if any step bypasses the new parsing/prettifying order.
> 
> **Overview**
> **Review metadata now uses parsed values** (API-shaped) when calling `prettifyFormValues` in `Onboarding`, `ContractorOnboarding`, `CreateCompany`, and `CostCalculator`, preventing mismatches between what is submitted and what is shown in the review/meta UI.
> 
> **Money prettification is corrected**: `prettifyFormValues` now converts `money` values *from cents* via `convertFromCents`, supports `null`, and adds an option `skipMoneyConversion` used when building review metadata from initial (already-normalized) values during auto-navigation to review.
> 
> **UI/test updates**: review step rendering is adjusted to hide empty `engagement_agreement_details`, introduces a shared `PrettifiedValuesRenderer` for tests, and adds/updates tests to assert correct money display in review (including direct-to-review invited employments).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e368313834b63d2ebd51cb9ca8258042894010e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/example/src/ReviewOnboardingStep.tsx
+++ b/example/src/ReviewOnboardingStep.tsx
@@ -243,17 +243,23 @@ export const ReviewOnboardingStep = ({
       >
         Edit Basic Information
       </button>
-      <h2 className='title'>Engagement Agreement Details</h2>
-      <ReviewMeta
-        meta={onboardingBag.meta.fields.engagement_agreement_details}
-      />
-      <button
-        className='back-button'
-        onClick={() => onboardingBag.goTo('engagement_agreement_details')}
-        disabled={onboardingBag.isEmploymentReadOnly}
-      >
-        Edit Engagement Agreement Details
-      </button>
+      {Object.keys(onboardingBag.meta.fields.engagement_agreement_details)
+        .length > 0 && (
+        <>
+          <h2 className='title'>Engagement Agreement Details</h2>
+          <ReviewMeta
+            meta={onboardingBag.meta.fields.engagement_agreement_details}
+          />
+          <button
+            className='back-button'
+            onClick={() => onboardingBag.goTo('engagement_agreement_details')}
+            disabled={onboardingBag.isEmploymentReadOnly}
+          >
+            Edit Engagement Agreement Details
+          </button>
+        </>
+      )}
+
       <h2 className='title'>Contract Details</h2>
       <ReviewMeta meta={onboardingBag.meta.fields.contract_details} />
       <button
--- a/src/components/form/utils.ts
+++ b/src/components/form/utils.ts
@@ -248,14 +248,6 @@ export const fieldTypesTransformations: Record<string, $TSFixMe> = {
       convertFromCents(value) ?? '',
     transformValueToAPI: () => convertToCents,
   },
-  [supportedTypes.RADIO]: {
-    transformValueToAPI: (field: $TSFixMe) => (value: string) => {
-      if (field.transformToBool) {
-        return value === 'yes';
-      }
-      return value;
-    },
-  },
   [supportedTypes.CHECKBOX]: {
     transformValueToAPI: (field: $TSFixMe) => (value: string | boolean) => {
       if (value === undefined) {
--- a/src/flows/ContractorOnboarding/hooks.tsx
+++ b/src/flows/ContractorOnboarding/hooks.tsx
@@ -811,26 +811,32 @@ export const useContractorOnboarding = ({
         select_country: prettifyFormValues(
           selectCountryInitialValues,
           stepFields.select_country,
+          { skipMoneyConversion: true },
         ),
         basic_information: prettifyFormValues(
           basicInformationInitialValues,
           stepFields.basic_information,
+          { skipMoneyConversion: true },
         ),
         contract_details: prettifyFormValues(
           contractDetailsInitialValues,
           stepFields.contract_details,
+          { skipMoneyConversion: true },
         ),
         contract_preview: prettifyFormValues(
           contractPreviewInitialValues,
           stepFields.contract_preview,
+          { skipMoneyConversion: true },
         ),
         pricing_plan: prettifyFormValues(
           pricingPlanInitialValues,
           stepFields.pricing_plan,
+          { skipMoneyConversion: true },
         ),
         eligibility_questionnaire: prettifyFormValues(
           eligibilityQuestionnaireInitialValues,
           stepFields.eligibility_questionnaire,
+          { skipMoneyConversion: true },
         ),
       };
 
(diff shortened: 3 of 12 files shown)


src/lib/utils.ts
```diff
@@ -230,10 +233,16 @@ export function prettifyFormValues(
...
+          // we convert to normal format and not use formatCurrency as consumers are expecting normal number
           return [
             key,
             {
-              prettyValue: value,
+              prettyValue:
+                !options?.skipMoneyConversion &&
+                (typeof value === 'string' || typeof value === 'number')
+                  ? convertFromCents(value)
+                  : value,
```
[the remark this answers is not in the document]
gabrielseco: it doesn't apply to our use case here

We get the value / options correctly

image

example/src/ReviewOnboardingStep.tsx
```diff
@@ -243,17 +243,23 @@ export const ReviewOnboardingStep = ({
...
-        meta={onboardingBag.meta.fields.engagement_agreement_details}
-      />
-      <button
-        className='back-button'
-        onClick={() => onboardingBag.goTo('engagement_agreement_details')}
-        disabled={onboardingBag.isEmploymentReadOnly}
-      >
-        Edit Engagement Agreement Details
-      </button>
+      {Object.keys(onboardingBag.meta.fields.engagement_agreement_details)
```
gabrielseco, line 246: this step is optional, we don't show it if we don't have any values to show

src/components/form/utils.ts
```diff
@@ -248,14 +248,6 @@ export const fieldTypesTransformations: Record<string, $TSFixMe> = {
       convertFromCents(value) ?? '',
     transformValueToAPI: () => convertToCents,
   },
-  [supportedTypes.RADIO]: {
```
gabrielseco, line 251: Bugbot was making a comment based on this

https://github.com/remoteoss/remote-flows/pull/1001#discussion_r3195489508

Basically it was me who added it on the past this but transformToBool isn't used

https://github.com/remoteoss/remote-flows/commit/ea796893c6009835564939a3564956813f6ff47d#diff-017c64080320946c917714b2f0e28fd3b2060379509d0fc465409480f7676365L128 -> this where I added it but as we aren't using anymore we can delete it

src/flows/ContractorOnboarding/tests/ContractorOnboarding.test.tsx
```diff
@@ -48,44 +48,12 @@ import {
 } from '@/src/common/api/fixtures/contractors-subscriptions';
 import { mockBlockedEligibilityQuestionnaireResponse } from '@/src/common/api/fixtures/eligibility-questionnaire';
 import { mockContractorBasicInformationSchema } from '@/src/common/api/fixtures/contractors';
+import { PrettifiedValuesRenderer } from '@/src/tests/components/PrettifiedValuesRenderer';
```
gabrielseco, line 51: made a shared component for the tests to avoid duplication

src/flows/ContractorOnboarding/hooks.tsx
```diff
@@ -953,12 +959,13 @@ export const useContractorOnboarding = ({
 
   async function onSubmit(values: FieldValues) {
     const currentStepName = stepState.currentStep.name;
+    const parsedValues = await parseFormValues(values);
+
     if (currentStepName in fieldsMetaRef.current) {
       fieldsMetaRef.current[
         currentStepName as keyof typeof fieldsMetaRef.current
-      ] = prettifyFormValues(values, stepFields[currentStepName]);
+      ] = prettifyFormValues(parsedValues, stepFields[currentStepName]);
```
gabrielseco, line 967: the bug that was happening before is that we were passing the values as we got them from the form.

For example for a money field the number was in correct format but for other fields like force field values containing integers the value is saved in cents

What happens here then?

When we submit, we invoke `parseFormValues` which transform all the money fields to cents including `annual_gross_salary`

then in prettifyFormValues we'll use the function called `convertFromCents` to make everything work

src/flows/ContractorOnboarding/hooks.tsx
```diff
@@ -811,26 +811,32 @@ export const useContractorOnboarding = ({
...
           stepFields.select_country,
+          { skipMoneyConversion: true },
         ),
         basic_information: prettifyFormValues(
           basicInformationInitialValues,
           stepFields.basic_information,
+          { skipMoneyConversion: true },
         ),
         contract_details: prettifyFormValues(
           contractDetailsInitialValues,
```
gabrielseco, line 822: here we build escape hatch to avoid transforming the values from cents to normal value

Why?

The reason is contractDetailsInitialValues already does this parsing for us, no need to do the conversion again.

Which also help us to make the change smaller as possible

src/flows/CostCalculator/CostCalculatorForm.tsx
```diff
@@ -107,7 +107,7 @@ export function CostCalculatorForm({
 
       if (costCalculatorBag?.meta?.fields) {
         costCalculatorBag.meta.fields = prettifyFormValues(
-          values,
+          parsedValues,
```
gabrielseco, line 110: I believe no one is using `costCalculatorBag.meta.fields` but it would make sense we pass parsedValues instead of the raw values

src/flows/Onboarding/tests/OnboardingFlow.test.tsx
```diff
@@ -153,15 +122,17 @@ describe('OnboardingFlow', () => {
...
           <div className='onboarding-review'>
             <h2 className='title'>Basic Information</h2>
-            <Review
-              values={onboardingBag.stepState.values?.basic_information || {}}
+            <PrettifiedValuesRenderer
+              values={onboardingBag.meta.fields.basic_information || {}}
             />
             <h2 className='title'>Contract Details</h2>
-            <Review
-              values={onboardingBag.stepState.values?.contract_details || {}}
```
gabrielseco, line 161: there was a bug in the test we shouldn't use stepState, if you check ReviewOnboardingStep.tsx in the example folder you'll see we use meta

cammellos (approved):

PR #1003, opened 2026-05-11
feat(review-step) - if initiated move to review step

gabrielseco:
## Problem

When the employment returns initiated which could happen after you invite the user, we don't move the user to the review step in both contractor management && onboarding flow

if we don't do this, when users try to edit an employment the API will fail as initiated definition is

```
- `initiated`: The user has been invited but has not completed registration.
```

## Solution

For both flows, we put the initiated status on the utils constants to make this work

## How to test

From what we've seen if the user.email is already on the system in my case ggarciaseco@gmail.com, in partners is where the initiated happened

## Loom
https://www.loom.com/share/60b799d9e668451e99d904091b4c5f71

---

> [!NOTE]
> **Medium Risk**
> Changes onboarding/contractor flows to treat the `initiated` employment status as read-only and disable invites, which affects core navigation and invite behavior across multiple flows but is localized to status gating logic.
> 
> **Overview**
> Ensures employments in `initiated` state are treated like other read-only statuses by adding `initiated` to the review-step allowed list in both `Onboarding` and `ContractorOnboarding`, so users are auto-routed to the review step instead of being allowed to edit.
> 
> Updates invite behavior to **prevent sending invites** when status is `initiated`: `OnboardingInvite` now disables its button based on `onboardingBag.canInvite` (and `initiated` is included in `disabledInviteButtonEmploymentStatus`), and the example `ReviewOnboardingStep` no longer passes/uses a separate `canInvite` prop.
> 
> Tests are expanded/adjusted to cover `initiated` for auto-navigation and to assert invite buttons are disabled for `initiated`, with additional waits to avoid timing flakiness around `canInvite` resolution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78c5faa15ce2dbde2c9218cc2b4529278dd5ac39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


example/src/ReviewOnboardingStep.tsx
```diff
@@ -167,20 +167,17 @@ export const MyOnboardingInviteButton = ({
   creditRiskStatus,
   Component,
   setErrors,
-  canInvite,
```
gabrielseco, line 170: I removed the canInvite here, consumers can still use it but in the ContractorOnboarding/OnboardingInvite we did the same use the canInvite internally that way they don't have to care if they can invite the user or not

src/flows/ContractorOnboarding/utils.ts
```diff
@@ -156,12 +156,14 @@ export const shouldIncludeProduct = (
  * @constant
  */
 export const reviewStepAllowedEmploymentStatus: Employment['status'][] = [
+  'initiated',
```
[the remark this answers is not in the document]
gabrielseco: For me its a false negative.

the initiated status is a final status that happens when you invite the employee or contractor.

Can the bug happen if the contractor doesn't have contract documents, well yes but that shouldn't happen as contract documents are created when you do the flow.

src/flows/Onboarding/tests/OnboardingInvite.test.tsx
```diff
@@ -287,7 +287,13 @@ describe('OnboardingInvite', () => {
 
     render(<OnboardingFlow {...defaultProps} />, { wrapper: TestProviders });
     const button = await screen.findByTestId('onboarding-invite');
+
+    await waitFor(() => {
```
gabrielseco, line 291: as we now have the canInvite we need to wait for the button as the canInvite depends on the employment http call

src/flows/Onboarding/tests/OnboardingInvite.test.tsx
```diff
@@ -1101,42 +1145,6 @@ describe('OnboardingInvite', () => {
       expect(MockCustomButton).not.toHaveBeenCalled();
     });
 
-    it('should pass through all props to custom button', async () => {
```
gabrielseco, line 1104: repeated test, removed

src/flows/Onboarding/utils.ts
```diff
@@ -100,12 +100,14 @@ export function buildSteps(config: StepConfig = {}) {
  * @constant
  */
 export const reviewStepAllowedEmploymentStatus: Employment['status'][] = [
+  'initiated',
```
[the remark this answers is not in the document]
gabrielseco: done

jordividaller (approved):

PR #1076, opened 2026-06-09
feat(gp): GP SDK foundation — shared hooks and flow scaffolding

hamzaremote:
## Summary

- Adds `useGPLegalEntities(companyId)` — public hook that fetches GP-enabled legal entities for a company
- Adds `useGPOnboardingSteps(employmentId)` — internal shared hook used by both GP flows
- Scaffolds `PayrollAdminOnboardingFlow` and `PayrollEmployeeOnboardingFlow` with types, context, hooks, and flow containers (step components are stubs; real implementations land in PBYR-4044 and PBYR-4045)
- Wires both flows into `src/index.tsx` and `package.json` exports map

Closes PBYR-4043 · Part of PBYR-3935

## Test plan

- [x] `npm run type-check` passes
- [x] `npm test -- --run` passes (781 tests)
- [x] `npm run build` produces `dist/flows/PayrollAdminOnboarding/` and `dist/flows/PayrollEmployeeOnboarding/`
- [x] Both flows and `useGPLegalEntities` are importable from `@remoteoss/remote-flows`

### DIFF
--- a/src/flows/PayrollAdminOnboarding/context.ts
+++ b/src/flows/PayrollAdminOnboarding/context.ts
@@ -1,0 +1,23 @@
+import type { usePayrollAdminOnboarding } from '@/src/flows/PayrollAdminOnboarding/hooks';
+import { createContext, useContext } from 'react';
+
+export const PayrollAdminOnboardingContext = createContext<{
+  formId: string | undefined;
+  adminBag: ReturnType<typeof usePayrollAdminOnboarding> | null;
+}>({
+  formId: undefined,
+  adminBag: null,
+});
+
+export const usePayrollAdminOnboardingContext = () => {
+  const context = useContext(PayrollAdminOnboardingContext);
+  if (!context.formId || !context.adminBag) {
+    throw new Error(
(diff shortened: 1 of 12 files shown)


src/flows/PayrollAdminOnboarding/context.ts
```diff
@@ -0,0 +1,23 @@
...
+});
+
+export const usePayrollAdminOnboardingContext = () => {
+  const context = useContext(PayrollAdminOnboardingContext);
+  if (!context.formId || !context.adminBag) {
+    throw new Error(
+      'usePayrollAdminOnboardingContext must be used within a PayrollAdminOnboardingFlow',
+    );
+  }
+  return context as {
```
jordividaller, line 19: Maybe you should use as const instead of redefining the type as it is done in the others context, but non-blocking.

jordividaller (approved):

src/flows/PayrollAdminOnboarding/context.ts
hamzaremote: ok if you don't mind I will handle it in a follow up MR

PR #1077, opened 2026-06-09
feat(gp): Global payroll demo flow

hamzaremote:
## Summary

- Implements the full `PayrollAdminOnboardingFlow` for Global Payroll admin onboarding
- Step sequence: Select Country → Contract Details → Administrative Details → Invite
- Employment created via `POST /v1/employments` with `type: global_payroll_employee`
- Subsequent steps use `PUT /v2/employments/:id/contract_details` and `PUT /v2/employments/:id/administrative_details`
- All mutations via `mutateAsyncOrThrow` (no deprecated `mutateAsync`)
- `useStepSubmitHandler` hook extracted to eliminate duplicated try/catch/navigate logic across form step components
- Example app page added under "GP Admin Onboarding" using the same styles as the EOR Onboarding demo

Closes PBYR-4044 · Part of PBYR-3935

## Test plan

- [x] `npm run type-check` passes
- [x] `npm test -- --run` passes (781 tests)
- [x] Lint: 0 errors
- [x] `npm run check-format` clean
- [x] Example app: "GP Admin Onboarding" tab shows country dropdown (225 countries), step nav bar, and card layout matching EOR Onboarding style
- [x] End-to-end: select a GP-supported country → schema loads → fill basic info → create employment → fill contract details → administrative details → send invitation (requires staging/partners env with GP schemas seeded)

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth/JWT token minting and proxy routing for employee APIs (dev-only guards on token routes), plus large new payroll onboarding surfaces that call employment and tax endpoints.
> 
> **Overview**
> This PR **replaces the GP onboarding scaffolds** with working **admin** and **employee** flows: JSON-schema–driven steps, per-step API mutations, shared submit/error handling, and typed step callbacks via `GPStepCallbacks`.
> 
> **Payroll admin** adds country selection, contract/administrative forms, employment creation (`POST /v1/employments`), updates, and an invite step wired through new API hooks and step components (no more null placeholders).
> 
> **Payroll employee** adds personal details, address, optional bank account, and **USA federal/state tax** steps with `taxStepsAvailability` (pending enrollment, jurisdiction, schema errors) and `countryCode` / `jurisdiction` on the flow props.
> 
> The **example app** registers GP Admin and GP Employee demos and extends the dev **proxy** to mint **employment-scoped JWT-bearer** tokens for `/v1/employee/*` using `x-rf-employment-id`, plus `authType: 'none'` when the proxy holds auth server-side. Bundle size limit is bumped slightly; lockfile has routine rollup-related updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ca1e3af231348462b70c25f9aed9118027cdcff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


example/src/App.tsx
```diff
@@ -175,6 +176,13 @@ const additionalDemos = [
     component: JsonSchemaComparisonDemo,
     sourceCode: JsonSchemaComparisonCode,
   },
+  {
```
jordividaller, line 180: I think we could put it upper in the menu just after contract onboarding

jordividaller (approved):

example/src/App.tsx
remotecom: I will do this in a followup MR

example/src/App.tsx
```diff
@@ -175,6 +177,20 @@ const additionalDemos = [
...
+    id: 'payroll-admin-onboarding',
+    title: 'GP Admin Onboarding',
+    description: 'Global Payroll admin onboarding flow (PBYR-4044)',
```
gabrielseco: nitpick: maybe remove the PBYR info

example/src/App.tsx
```diff
@@ -175,6 +177,20 @@ const additionalDemos = [
...
+    component: PayrollAdminOnboardingForm,
+    sourceCode: '',
+  },
+  {
+    id: 'payroll-employee-onboarding',
+    title: 'GP Employee Onboarding',
+    description: 'Global Payroll employee self-onboarding flow (PBYR-4045)',
```
gabrielseco: nitpick: maybe remove the PBYR info

example/src/PayrollAdminOnboarding.tsx
```diff
@@ -0,0 +1,274 @@
...
+              </ul>
+            </div>
+
+            <div className='card' style={{ marginBottom: 20 }}>
+              <h1 className='heading'>{STEP_LABELS[currentStep]}</h1>
+              <p style={{ fontSize: 14, color: '#71717A', marginBottom: 24 }}>
+                {STEP_DESCRIPTIONS[currentStep]}
+              </p>
+
+              {currentStep === 'select_country' && (
```
gabrielseco: nitpick: probably its better to do a switch like in the other flows but don't have a strong preference

example/src/PayrollAdminOnboarding.tsx
```diff
@@ -0,0 +1,274 @@
...
+                        e.fieldErrors.map((fe) => ({
+                          ...fe,
+                          userFriendlyLabel: fe.field,
+                        })),
+                      )
+                    }
+                    onSuccess={clearErrors}
+                  />
+                  <AlertError errors={errors} />
+                  {adminBag.countryCode && adminBag.fields.length > 0 && (
```
gabrielseco: these ifs are strange, we don't have these for other demos why?

example/src/PayrollAdminOnboarding.tsx
```diff
@@ -0,0 +1,274 @@
...
+                  <AlertError errors={errors} />
+                  <div className='buttons-container'>
+                    <BackButton className='back-button' onClick={clearErrors}>
+                      Previous Step
+                    </BackButton>
+                    <SubmitButton
+                      className='submit-button'
+                      onClick={clearErrors}
+                    >
+                      Save &amp; Continue
```
gabrielseco: nitpick: you can do &

example/src/PayrollAdminOnboarding.tsx
```diff
@@ -0,0 +1,274 @@
...
+            </div>
+          </>
+        );
+      }}
+    />
+  );
+}
+
+function GPAdminOnboardingInner() {
+  const { data: legalEntities, isLoading } = useGPLegalEntities(COMPANY_ID);
```
gabrielseco, line 230: Question: why this is not done internally and you pass the companyId?

is companyId setup by the partner and we cannot use an identity endpoint?

src/components/form/fields/default/DatePickerFieldDefault.tsx
```diff
@@ -62,14 +66,21 @@ export function DatePickerFieldDefault({
         >
           <Calendar
             mode='single'
+            captionLayout='dropdown'
```
gabrielseco, line 72: non-important: I don't know if I like this from what I saw in Onboarding

Image

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -46,10 +57,11 @@ export const usePayrollAdminOnboarding = ({
     string | undefined
   >(initialCountryCode);
 
-  // Fix: derive from state, not from the prop, so setInternalCountryCode changes are reflected
-  const skipCountry = !!internalCountryCode;
+  const [fieldValues, setFieldValues] = useState<FieldValues>({});
```
gabrielseco: fieldValues already is on the `useStepState` why redeclare it?

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -66,6 +78,77 @@ export const usePayrollAdminOnboarding = ({
...
+    'global_payroll_administrative_details',
+    fieldValues,
+    {
+      enabled:
+        currentStep === 'administrative_details' && !!internalCountryCode,
+    },
+  );
+
+  // Current step's schema form
+  const currentSchema = useMemo(() => {
```
gabrielseco, line 116: this would look better in a object or map

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -76,20 +159,128 @@ export const usePayrollAdminOnboarding = ({
...
+    setFieldValues,
+    // Form helpers
+    handleValidation,
+    parseFormValues,
+    onSubmit,
+    sendInvite,
+    // Navigation
+    setInternalCountryCode,
+    setInternalEmploymentId,
     goToNextStep: nextStep,
```
gabrielseco: I'd rename next and back to follow conventions from other flows

src/flows/PayrollAdminOnboarding/components/InvitationStep.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { components } = useFormFields();
+
+  const CustomButton = components?.button;
+  if (!CustomButton) {
+    throw new Error('Button component not found');
+  }
+
+  const handleInvite = async () => {
+    try {
+      const data = await adminBag.sendInvite();
```
gabrielseco: I think it would be better to not have sendInvite and just call the mutation directly like in other components

src/flows/PayrollAdminOnboarding/components/SelectCountryStep.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export function SelectCountryStep({
+  onSubmit,
+  onSuccess,
+  onError,
+}: GPStepCallbacks) {
+  const { adminBag } = usePayrollAdminOnboardingContext();
+  const { client } = useClient();
+
+  const { data: countriesResponse, isLoading: isLoadingCountries } = useQuery(
```
gabrielseco: this is strange in other forms we create a custom json schema to render the form

src/flows/PayrollAdminOnboarding/components/SelectCountryStep.tsx
```diff
@@ -0,0 +1,91 @@
...
+          error: error as Error,
+          rawError: error as Record<string, unknown>,
+          fieldErrors: [],
+        });
+      }
+    }
+  };
+
+  return (
+    <div className='space-y-4'>
```
gabrielseco: this code means partners won't be able to override this step

src/flows/types.ts
```diff
@@ -112,3 +113,16 @@ export type JSONSchemaFormResultWithFieldsets = FormResult & {
     'x-jsf-presentation'?: Record<string, unknown>;
   };
 };
+
+/**
+ * Shared callback prop type for GP step components (admin and employee flows).
+ */
+export type GPStepCallbacks = {
```
gabrielseco: non-important: I usually create exact types of what I'll be returning instead of planting a generic

src/flows/PayrollEmployeeOnboarding/tests/components.test.tsx
```diff
@@ -0,0 +1,299 @@
...
+import { BankAccountStep } from '../components/BankAccountStep';
+import { FederalTaxesStep } from '../components/FederalTaxesStep';
+import { HomeAddressStep } from '../components/HomeAddressStep';
+import { PersonalDetailsStep } from '../components/PersonalDetailsStep';
+import { StateTaxesStep } from '../components/StateTaxesStep';
+import { SubmitButton } from '../components/SubmitButton';
+import { useEmployeeStepSubmitHandler } from '../components/useEmployeeStepSubmitHandler';
+import { usePayrollEmployeeOnboardingContext } from '../context';
+import { useFormFields } from '@/src/context';
+import { ButtonDefault } from '@/src/components/form/fields/default/ButtonDefault';
```
gabrielseco, line 13: we usually create a test for each file for example SubmitButton, BackButton etc...

also I believe we're mocking to much, I see too many unit testing here instead of testing the flow with msw like we do for other files

src/flows/PayrollEmployeeOnboarding/components/StateTaxesStep.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * Render only when `employeeBag.taxStepsAvailability.state_taxes.isAvailable`
+ * is true (USA + jurisdiction set + post-enrollment). Returns null otherwise.
+ * Submits to PUT /v1/employee/state-taxes/{jurisdiction} where jurisdiction
+ * comes from the flow's `jurisdiction` prop.
+ */
+export function StateTaxesStep(props: GPStepCallbacks) {
+  const { employeeBag } = usePayrollEmployeeOnboardingContext();
+  const handleSubmit = useEmployeeStepSubmitHandler(props);
+
+  if (!employeeBag.taxStepsAvailability.state_taxes.isAvailable) return null;
```
gabrielseco, line 16: Question: I haven't tested but you could land in the step nothing gets rendered, shouldn't it better to not include the step when this happens?

src/flows/PayrollEmployeeOnboarding/components/FederalTaxesStep.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * Render only when `employeeBag.taxStepsAvailability.federal_taxes.isAvailable`
+ * is true. Returns null otherwise so the consumer can render their own
+ * not-available UI driven by `unavailableReason`. The step also flips itself
+ * to unavailable retroactively if the backend returns 404 on submit.
+ */
+export function FederalTaxesStep(props: GPStepCallbacks) {
+  const { employeeBag } = usePayrollEmployeeOnboardingContext();
+  const handleSubmit = useEmployeeStepSubmitHandler(props);
+
+  if (!employeeBag.taxStepsAvailability.federal_taxes.isAvailable) return null;
```
gabrielseco, line 16: Question: I haven't tested but you could land in the step nothing gets rendered, shouldn't it better to not include the step when this happens?

src/flows/PayrollEmployeeOnboarding/components/BankAccountStep.tsx
```diff
@@ -0,0 +1,29 @@
...
+ */
+export function BankAccountStep(props: GPStepCallbacks) {
+  const { employeeBag } = usePayrollEmployeeOnboardingContext();
+  const handleSubmit = useEmployeeStepSubmitHandler(props);
+
+  const isRequired = employeeBag.selfOnboardingSubsteps.some(
+    (s) => s.type === 'employee_provides_bank_details',
+  );
+
+  if (!isRequired) return null;
```
gabrielseco, line 19: Question: I haven't tested but you could land in the step nothing gets rendered, shouldn't it better to not include the step when this happens?

example/src/PayrollEmployeeOnboarding.tsx
```diff
@@ -0,0 +1,549 @@
...
+          change
+        </button>
+      </p>
+      <EmployeeFlowForm employmentId={employmentId} />
+    </div>
+  );
+}
+
+export function PayrollEmployeeOnboardingForm() {
+  return (
```
gabrielseco, line 536: I see too much code here that partners will need to reimplement on their own

src/flows/PayrollEmployeeOnboarding/hooks.tsx
```diff
@@ -1,29 +1,80 @@
...
+type TaxStepKey = (typeof TAX_STEPS)[number];
+
 export const usePayrollEmployeeOnboarding = ({
   employmentId,
+  countryCode,
+  jurisdiction,
   initialValues,
   options,
 }: Omit<PayrollEmployeeOnboardingFlowProps, 'render'>) => {
+  const [fieldValues, setFieldValues] = useState<FieldValues>({});
```
gabrielseco: same question why an additional state fieldValues is on the useStepState

src/flows/PayrollEmployeeOnboarding/hooks.tsx
```diff
@@ -53,16 +108,301 @@ export const usePayrollEmployeeOnboarding = ({
...
+    {
+      enabled:
+        isUSA &&
+        !!jurisdiction &&
+        isPostEnrollment &&
+        currentStep === 'state_taxes',
+    },
+  );
+
+  const currentSchema = useMemo(() => {
```
gabrielseco, line 228: i believe an object or a map would be better

src/flows/PayrollEmployeeOnboarding/hooks.tsx
```diff
@@ -53,16 +108,301 @@ export const usePayrollEmployeeOnboarding = ({
...
+    federalTaxesSchema.data,
+    stateTaxesSchema.data,
+  ]);
+
+  // Availability is computed AFTER schema queries so we can fold their error
+  // state (e.g. backend returns 400 for an unseeded schema) into a friendly
+  // `schema_unavailable` reason instead of letting the consumer render an
+  // empty form.
+  const taxStepsAvailability = useMemo(() => {
+    const federalReason: TaxStepUnavailableReason | null = !isUSA
```
gabrielseco: change this please, nested ternaries are bad for eligibility

gabrielseco (commented):
there are too many comments

src/flows/PayrollAdminOnboarding/hooks.tsx
hamzaremote: Done in aa456a39 — pulled `fieldValues`/`setFieldValues` from `useStepState` instead of declaring our own, same change applied in the employee hook.

src/flows/PayrollAdminOnboarding/hooks.tsx
hamzaremote: Done in aa456a39 — replaced both step→schema lookups with a `Partial<Record<StepKey, …>>` map (admin + employee).

src/flows/types.ts
hamzaremote: Done in aa456a39 — `GPStepCallbacks<TSuccess = unknown>` is now generic; default stays `unknown` so existing callers keep compiling, and any step can specialise its onSuccess payload.

example/src/App.tsx
hamzaremote: Done in aa456a39 — dropped the PBYR codes.

src/flows/PayrollEmployeeOnboarding/hooks.tsx
hamzaremote: Done in aa456a39 — both branches rewritten as early-return IIFEs (`!isUSA` → `unsupported_country`, schema/enrollment checks below).

src/flows/PayrollAdminOnboarding/hooks.tsx
hamzaremote: Done in aa456a39 — renamed to `next`/`back` across both flows and all internal call sites.

example/src/PayrollAdminOnboarding.tsx
hamzaremote: Done in aa456a39 — switched to a `renderStep` switch matching `Onboarding.tsx`, and factored a shared `onStepError` helper to cut the per-case boilerplate.

example/src/PayrollAdminOnboarding.tsx
hamzaremote: Done in aa456a39 — using `&` directly in JSX text now.

src/flows/PayrollEmployeeOnboarding/hooks.tsx
hamzaremote: Done in aa456a39.

src/flows/PayrollAdminOnboarding/components/InvitationStep.tsx
hamzaremote: Done in aa456a39 — `InvitationStep` now calls `useGPInviteEmployee` directly and refetches the steps itself; `sendInvite` removed from the admin hook surface.

src/flows/PayrollEmployeeOnboarding/hooks.tsx
hamzaremote: Done in aa456a39 — same fix as the admin hook.

example/src/App.tsx
hamzaremote: Done in aa456a39.

src/flows/PayrollEmployeeOnboarding/components/FederalTaxesStep.tsx
hamzaremote: Same fix as the state-taxes thread — visibility computed from `isUSA && isPostEnrollment` and the step is dropped from `next()`/`back()` when those don't hold.

src/components/form/fields/default/DatePickerFieldDefault.tsx
hamzaremote: Fair — I added the dropdown caption so e.g. `date_of_birth` (year range 1920-2026) doesn't force users to click 100+ times. Onboarding's date picker also benefits but doesn't enable it because its only date field is hire_date with a narrow range. Happy to gate this behind a prop or feature-flag the dropdown caption if you'd rather not change the default — let me know which you'd prefer.

example/src/PayrollEmployeeOnboarding.tsx
hamzaremote: Yes — there's room to pull a lot of the demo wiring (two-context auth, the `fetch-employee-token` route, the schema-aware default-values plumbing) into the SDK so partners don't reimplement it. I want to do that in a dedicated MR rather than expanding this one — tracking it.

example/src/PayrollAdminOnboarding.tsx
hamzaremote: These guards exist because `select_country` rolls two actions into one step: (1) pick a country, (2) fill the country-specific basic-information form and submit to create the employment. Until both `countryCode` is set AND the schema's fields have loaded, there's nothing to submit. In aa456a39 I extracted it as `const isFormReady = !!adminBag.countryCode && adminBag.fields.length > 0` so the intent reads better. Splitting select_country into two steps (like `Onboarding` does) would let us drop the guard entirely — happy to do that in a follow-up MR if you'd prefer that shape.

src/flows/PayrollEmployeeOnboarding/tests/components.test.tsx
hamzaremote: Acknowledged — the test file as-is is more a coverage-floor than a proper test surface. I'd like to keep this MR scoped and follow up with: (1) split per-component files, (2) MSW for the schema/PUT flow, and (3) drop the manual mocks. Opening a ticket.

example/src/PayrollAdminOnboarding.tsx
hamzaremote: Partners always have their own `companyId` from the integration setup, so the prop forwards it from there. That said, `/v1/identity/current` returns `data.company` for company-scoped tokens, so we could fetch it internally and drop the prop. I'd rather do that in a follow-up so this MR stays scoped to the admin flow — opening a ticket for it.

src/flows/PayrollAdminOnboarding/components/SelectCountryStep.tsx
hamzaremote: Tied to the same point — once we move SelectCountryStep to a fully-jsf-driven shape, partners override it the same way they override any other step (via the `components` prop / by rendering their own `<SelectCountryStep>` replacement). Tracked in the same follow-up.

src/flows/PayrollEmployeeOnboarding/components/StateTaxesStep.tsx
hamzaremote: Good catch — in aa456a39 the steps now carry a `visible` flag computed from the same signals (`isUSA && isPostEnrollment` for federal, plus `!!jurisdiction` for state, and `employee_provides_bank_details` substep for bank). `next()`/`back()` skip hidden steps, so the user can't land on an empty one through the navigation. The `return null` in each step body stays as a safety net for partners that render the step directly.

src/flows/PayrollAdminOnboarding/components/SelectCountryStep.tsx
hamzaremote: You're right that the rest of the flow uses a single json schema. The wrinkle here is that the basic-information schema is country-specific — until the user picks a country we don't have a schema to render. So this step currently hand-rolls the `<select>` and only renders `PayrollAdminForm` once `countryCode` is set. A cleaner shape would be a small upfront jsf (country only) → on submit fetch the country-specific schema → render that as the second jsf, mirroring Onboarding's two-step flow. I'd prefer to ship this in a follow-up so this MR stays focused — opening a ticket.

src/flows/PayrollEmployeeOnboarding/components/BankAccountStep.tsx
hamzaremote: Same fix — bank-account step is hidden when `employee_provides_bank_details` isn't in the self-onboarding substeps, so navigation skips it.

gabrielseco (approved):

gabrielseco (dismissed):

PR #1111, opened 2026-06-24
feat: expose company open tasks flags on contractor onboarding review step

cammellos:
## Summary

- Adds \`needsCompanyVerification\` and \`needsRemotePaymentsSetup\` boolean flags to \`contractorOnboardingBag\`, sourced from the new \`GET /v1/companies/{company_id}/actions\` endpoint (backend MR [84912](https://gitlab.com/remote-com/employ-starbase/tiger/-/merge_requests/84912))
- The query only fires on the \`review\` step and requires a valid \`companyId\` from the identity
- Example app demonstrates both nudge banners on the review step, conditionally shown after a contractor is invited
- Manually adds \`getV1CompaniesCompanyIdActions\`, \`CompanyAction\`, and \`CompanyActionsResponse\` to the production-generated API client (rather than regenerating wholesale from the local backend, which strips \`security\` annotations and breaks auth on all existing endpoints)
- Fixes example app company-manager JWT scope to use \`all:write\`

## Dependencies

Depends on backend MR 84912 being merged and deployed before this PR can be merged.

## Test plan

- [ ] Load the contractor onboarding review step with a company that has \`verify_company\` open task → \`CompanyVerificationBanner\` appears after invite
- [ ] Load with a company that has \`setup_remote_payments\` open task → \`RemotePaymentServicesSetUp\` banner appears after invite
- [ ] Load with a company with no open tasks → neither banner appears
- [ ] \`npm run type-check\` passes
- [ ] \`npm run ci\` passes


src/client/sdk.gen.ts
```diff
@@ -3761,6 +3764,37 @@ export const getV1CompaniesCompanyId = <ThrowOnError extends boolean = false>(
...
+ *
+ *
+ * ## Scopes
+ *
+ * | Category | Read only Scope | Write only Scope (read access implicit) |
+ * |---|---|---|
+ * | Manage company resources (`company_admin`) | View companies (`company:read`) | Manage companies (`company:write`) |
+ *
+ */
+export const getV1CompaniesCompanyIdActions = <
```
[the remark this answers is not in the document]
jordividaller: BE MR are now merged, so we could generated from production directly to avoid to have manually changes.

example/api/jwt_auth.js
```diff
@@ -3,8 +3,7 @@ const jwt = require('jsonwebtoken');
 
 const { buildGatewayURL } = require('./utils.js');
 
-const DEFAULT_SCOPES =
-  'company_admin employments invoices time_and_attendance employment_documents employment_payments payroll';
+const DEFAULT_SCOPES = 'all:write';
```
jordividaller, line 6: non-blocking because it is for the example app but why don't adding only the new scope ?

cammellos: sorry, i don't understand the question :)

cammellos: oh I get it now, there's no reason to be selective with scopes, since it's just the example app, and that breaks if we add some that are not in there

jordividaller: Ok understood  thx!

jordividaller (approved):

PR #1142, opened 2026-07-03
feat(gp): PayrollAdminOnboarding flow — mutations, schemas, and step components (PBYR-4044)

hamzaremote:
## Summary

Completes the `PayrollAdminOnboarding` flow skeleton that was scaffolded in an earlier PR. The scaffold had step state and a render-prop container but no mutations, no schema loading, and no step components — this PR adds all of that.

Part of the GP payroll split plan (`GP_PAYROLL_SPLIT_PLAN.md`). Stacks on #1134.

**What's added:**

- **`api.ts`** — React Query hooks for the four admin API calls: `useGPFormSchema` (schema per step), `useGPCreateEmployment`, `useGPUpdateContractDetails`, `useGPUpdateAdministrativeDetails`, `useGPInviteEmployee`
- **`hooks.tsx`** — mutations wired into `onSubmit`, one schema query per step gated by `enabled: currentStep === X` (lazy, never fetches ahead), `refetchSteps()` called after each successful submit
- **`components/`** — `SelectCountryStep`, `ContractDetailsStep`, `AdministrativeDetailsStep`, `InvitationStep`, `PayrollAdminForm`, `SubmitButton`, `BackButton`, `useStepSubmitHandler`
- **`src/flows/types.ts`** — `GPStepCallbacks<TSuccess>` shared callback type (used here and by the employee flow in the next PR)
- **`src/index.tsx`** — exports `GPAdminStepCallbacks`, `useGPOnboardingSteps`

**Note on `InvitationStep`:** this step doesn't go through `useStepSubmitHandler` because there's no form — it's a single button that fires the invite mutation directly. It owns its mutation instance and calls `adminBag.refetchSteps()` / `adminBag.next()` itself.

## Test plan

- [ ] Run `npm run type-check` — no errors
- [ ] Run `npm run test` — passes
- [ ] Verify `npm run build` completes without OOM (heap raise from #1134 helps here)

---

> [!NOTE]
> **Medium Risk**
> New multi-step flow creates and mutates employments and sends invites via several API calls; incorrect skip/resume or country sync could leave the flow without a valid employment ID, though guards and update-vs-create logic mitigate that.
> 
> **Overview**
> Completes the **PayrollAdminOnboarding** scaffold by wiring country/step JSON schemas, employment mutations, and real step UI on top of the existing step-state hook.
> 
> **`api.ts`** adds React Query hooks for country-specific form schemas (with optional `employment_id` for role-aware fields), GP country picker schema, create employment, update basic/contract/admin details, and invite.
> 
> **`hooks.tsx`** now loads the right schema per step (country picker until a country is chosen, then basic information on `select_country`), validates/parses via JSF, and routes **`onSubmit`** to create or update employment data with **`refetchSteps()`** after each success. Country skip only applies when **both** `employmentId` and country are known (resume), avoiding a path with no employment ID.
> 
> **Step components** (`PayrollAdminForm`, country/contract/admin steps, submit/back, **`InvitationStep`** with its own invite mutation) use shared **`GPStepCallbacks`** and **`useStepSubmitHandler`** (server field errors mapped onto the form). Public exports include **`GPAdminStepCallbacks`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b5987135e57e70b6046531006ef78a5984b019f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/flows/PayrollAdminOnboarding/components/SelectCountryStep.tsx
jordividaller, line 1: We don't render views in SDK components, only the form logic. Everything else is provided by the JSON schema and is fully customizable by the client. We should do it the same way as in the other flows.

jordividaller (changes requested):
## General concern: the flow re-implements mechanics that already exist in the lib

Overall this flow tends to reinvent logic that other flows already share, instead of reusing it. A few concrete points:

**1. Error handling — reuse `handleStepError`.**
There's a shared helper [`handleStepError`](src/lib/utils.ts#L329) already used by 9 components across `ContractorOnboarding` (7) and `CreateCompany` (2). It does the `isMutationError` check + `normalizeFieldErrors` + maps errors back onto the form fields in one call. This flow ignores it and re-inlines the same `try/catch isMutationError` block **three times**:
- `src/flows/PayrollAdminOnboarding/components/useStepSubmitHandler.ts`
- `src/flows/PayrollAdminOnboarding/components/SelectCountryStep.tsx`
- `src/flows/PayrollAdminOnboarding/components/InvitationStep.tsx`

Please route these through `handleStepError` like the other flows.

**2. Field errors aren't normalized — likely a functional regression, not just style.**
The other flows pass `NormalizedFieldError[]` (via `handleStepError` / `normalizeFieldErrors`), which is what attaches server-side validation errors to the correct form fields. Here we pass the raw `error.fieldErrors` (`FieldError[]`) with no normalization, so field-level API errors probably won't render under their fields the way they do elsewhere. On top of that, the public `GPStepCallbacks` type freezes `fieldErrors: FieldError[]` while the rest of the public API exposes `NormalizedFieldError[]` — that's an inconsistent public contract. Suggest normalizing and aligning the type on `NormalizedFieldError[]`.

**3. `SelectCountryStep` hand-rolls the country field** instead of going through the schema/field pipeline (see the separate comment). `ContractorOnboarding` even reuses `selectCountryStepSchema` + `countriesOptions` for this.

### What's legitimately new (not a concern)

- **`useGPFormSchema`** is fair: unlike the other flows that co-locate static schemas in `json-schemas/`, GP needs a remote per-country/per-form schema (`getV1CountriesCountryCodeForm`), so a new hook is justified. Minor style nit only: the three parallel `useQuery` calls gated by `enabled` + the manual `currentSchema` switch could collapse into a single hook parameterized by step.
- **`PayrollAdminForm` / `useStepSubmitHandler`** are fine to exist — every flow has its own `<XxxForm>`, so that's the norm, not reinvention. The only issue is that `useStepSubmitHandler` doesn't call `handleStepError` (see point 1).

Net: the new remote-schema piece is good; the concern is only about the parts that already existed as shared helpers.

src/flows/PayrollAdminOnboarding/components/SelectCountryStep.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const { adminBag } = usePayrollAdminOnboardingContext();
+  const handleSubmit = useStepSubmitHandler(props);
+
+  const defaultValues =
+    (adminBag.stepState.values?.select_country as Record<string, unknown>) ||
+    (adminBag.initialValues?.basic_information as Record<string, unknown>);
+
+  return (
+    <PayrollAdminForm
+      key={adminBag.countryCode ?? 'no-country'}
```
gabrielseco, line 18: is the key to reset the form?

src/index.tsx
```diff
@@ -108,7 +109,10 @@ export type {
   PayrollEmployeeOnboardingRenderProps,
 } from '@/src/flows/PayrollEmployeeOnboarding';
 
-export { useGPLegalEntities } from '@/src/common/api/gpOnboarding';
+export {
+  useGPLegalEntities,
+  useGPOnboardingSteps,
```
gabrielseco: useGPOnboardingSteps I need to understand in a next PR how this is used as we tend to avoid it on example app to call directly the hook

if we're exposing inside the adminBag.apiSteps maybe we don't need to export it right?

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -44,12 +56,20 @@ export const usePayrollAdminOnboarding = ({
 
   const [internalCountryCode, setInternalCountryCode] = useState<
     string | undefined
-  >(initialCountryCode);
+  >(
+    initialCountryCode ??
```
gabrielseco: why not initialCountryCode instead of ((initialValues?.basic_information as Record<string, unknown>)
        ?.country_code as string | undefined),

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -63,8 +83,67 @@ export const usePayrollAdminOnboarding = ({
...
+
+  const currentStep = stepState.currentStep.name;
+
+  const schemaTypeByStep: Partial<Record<AdminStepKey, GPAdminSchemaType>> = {
+    select_country: 'global_payroll_basic_information',
+    contract_details: 'global_payroll_contract_details',
+    administrative_details: 'global_payroll_administrative_details',
+  };
+
+  // Always-on: provides the country picker schema for the select_country step
```
gabrielseco: I don't think the comment provides value, we just call the hook to initialize the static json schema

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -63,8 +83,67 @@ export const usePayrollAdminOnboarding = ({
...
+  // before a country has been chosen. Hooks must not be called conditionally.
+  const countrySelectSchemaQuery = useGPCountrySelectSchema(fieldValues);
+
+  const currentSchemaQuery = useGPFormSchema(
+    internalCountryCode,
+    schemaTypeByStep[currentStep],
+    fieldValues,
+    {
```
gabrielseco, line 112: wondering if this query shouldn't be only called when currentStep !== 'select_country' and that way we always pass internalEmploymentId

gabrielseco (approved):

src/flows/PayrollAdminOnboarding/components/SelectCountryStep.tsx
hamzaremote: Yes — the `key={adminBag.countryCode ?? 'no-country'}` remounts `PayrollAdminForm` when the country changes so it re-initializes against the newly loaded basic-information schema instead of the country-picker one. Added a comment making that explicit.

src/index.tsx
hamzaremote: Good call — removed the `useGPOnboardingSteps` public export. It's consumed only internally by the admin/employee flows and is already surfaced via `adminBag.apiSteps`, so the standalone export was redundant. (It was added in this PR, so dropping it isn't a breaking change to any released contract.)

src/flows/PayrollAdminOnboarding/hooks.tsx
hamzaremote: The `initialCountryCode` prop *is* used — it takes precedence, then we fall back to the country from `initialValues.basic_information` (for resume). Cleaned up the ugly inline cast into a typed `initialValuesCountryCode` variable: `initialCountryCode ?? initialValuesCountryCode`.

src/flows/PayrollAdminOnboarding/hooks.tsx
hamzaremote: Trimmed the comment down to a concise one-liner.

src/flows/PayrollAdminOnboarding/hooks.tsx
hamzaremote: Left the logic as-is on purpose: `useGPFormSchema` is already gated by `enabled: !!countryCode && !!schemaType`, so during the picker phase (no country yet) it never fetches. Gating on `currentStep !== 'select_country'` would actually break the case where a country is picked *while still on* the `select_country` step — we need the basic-information schema to load there. `employmentId` is already passed only when not on `select_country` via the ternary.

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -76,9 +154,121 @@ export const usePayrollAdminOnboarding = ({
...
+    },
+    [currentSchema],
+  );
+
+  const parseFormValues = useCallback(
+    async (values: FieldValues): Promise<Record<string, unknown>> => {
+      if (!currentSchema) return values;
+      return parseJSFToValidate(values, currentSchema.fields, {
+        isPartialValidation: false,
+      });
```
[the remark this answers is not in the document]
hamzaremote: Fixed in fa9f1a31. `SubmitButton` now also disables while `adminBag.isLoading` (schema/steps still loading), so a submit can no longer fire during the window where `handleValidation` returns null and `parseFormValues` passes raw values.

src/flows/PayrollAdminOnboarding/components/useStepSubmitHandler.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  return async (values: Record<string, unknown>) => {
+    try {
+      await onSubmit?.(values);
+      const data = await adminBag.onSubmit(values);
+      await onSuccess?.(data);
+      adminBag.next();
+    } catch (error: unknown) {
+      onError?.(handleStepError(error));
+    }
```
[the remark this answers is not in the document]
hamzaremote: Fixed in fa9f1a31. `PayrollAdminForm` now passes its React Hook Form instance into the submit handler (`onSubmit(values, form)`), and `useStepSubmitHandler` forwards it to `handleStepError(error, undefined, form)` so mutation `fieldErrors` are mapped back onto the matching form fields via `setFormFieldErrors` — same wiring as `ContractorOnboardingForm`. (Field metadata is left `undefined` because the GP admin `meta` only carries fieldsets, not a field-name→label map; labels fall back to the field name, matching existing best-effort behavior.)

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -76,9 +154,121 @@ export const usePayrollAdminOnboarding = ({
...
+          const empId = (data as { data?: { employment?: { id?: string } } })
+            ?.data?.employment?.id;
+          if (!empId) {
+            throw new Error(
+              'Employment was created but no ID was returned. Cannot proceed.',
+            );
+          }
+          setInternalEmploymentId(empId);
+          await refetchSteps();
+          return data;
```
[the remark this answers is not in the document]
hamzaremote: Fixed in 5d7a727b. `createEmployment` on `select_country` is now guarded behind `!internalEmploymentId`, so if a prior attempt already created the employment (e.g. a post-create step like `refetchSteps` failed and the user retried), the retry reuses the existing employment and just refetches/advances instead of creating a second one.

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -76,9 +154,121 @@ export const usePayrollAdminOnboarding = ({
...
+          if (!internalCountryCode) {
+            throw new Error(
+              'Country code is required to create an employment.',
+            );
+          }
+          const data = await createEmploymentAsync({
+            countryCode: internalCountryCode,
+            legalEntityId,
+            basicInformation: parsedValues,
+          });
```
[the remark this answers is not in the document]
hamzaremote: Fixed in 5d7a727b. The create now uses the country_code from the just-submitted (parsed) form values, falling back to `internalCountryCode` only if absent — so the employment is created with the same country the `basicInformation` was validated against, eliminating the mismatch when the `internalCountryCode` sync hasn't caught up.

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -76,9 +154,132 @@ export const usePayrollAdminOnboarding = ({
...
+              basicInformation: parsedValues,
+            });
+            const empId = (data as { data?: { employment?: { id?: string } } })
+              ?.data?.employment?.id;
+            if (!empId) {
+              throw new Error(
+                'Employment was created but no ID was returned. Cannot proceed.',
+              );
+            }
+            setInternalEmploymentId(empId);
```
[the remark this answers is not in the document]
hamzaremote: Fixed in 2b598713. The `select_country` handler now calls `setInternalCountryCode(countryCode)` with the resolved code before advancing, so the contract/administrative schemas — which are keyed on `internalCountryCode` — load correctly on the following steps.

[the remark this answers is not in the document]
hamzaremote: Fixed in 2b598713. Instead of skipping the write when `internalEmploymentId` is already set, the handler now updates basic information via `putV2EmploymentsEmploymentIdBasicInformation` (new `useGPUpdateBasicInformation` hook). So returning to this step and editing persists the changes, while still never creating a second employment.

jordividaller (approved):

PR #1171, opened 2026-07-17
feat(gp): example app demo for admin onboarding flow (PBYR-4044)

hamzaremote:
## What

The final piece of the Global Payroll onboarding work (see [split plan](GP_PAYROLL_SPLIT_PLAN.md), PR 5): the **example-app demos** for the admin and employee flows, plus the server-side auth plumbing they need. Both SDK flows (#1142, #1168) are already merged to `main`.

## Contents

- **`example/api/jwt_auth.js`** — server-side minting of an employee-assertion JWT-bearer token (`getEmployeeToken`), plus a production guard on the token endpoints.
- **`example/api/proxy.js`** — path-based token routing: `/v1/employee/*` → employee assertion (employment resolved from the `x-rf-employment-id` header, minted server-side so the FE never holds it); everything else unchanged.
- **`example/api/routes.js`** — wire the `/api/fetch-employee-token/:employmentId` route.
- **`example/src/RemoteFlows.tsx`** — add `authType: 'none'` so the inner (employee) context skips the FE auth callback while the proxy mints tokens.
- **`example/src/PayrollAdminOnboarding.tsx` / `PayrollEmployeeOnboarding.tsx`** — the two demos.
- **`example/src/App.tsx`** — register both demos.

## Design notes (the open questions from the arch doc)

- **Dual-`RemoteFlows`-provider pattern** (employee demo): the outer context uses a company-manager token to read `country`/`jurisdiction` from the employment (the employee assertion token can't fetch `/v1/employments/:id`); the inner context uses the employee-scoped token minted by the proxy.
- **Proxy auth contract**: the FE sends `x-rf-employment-id`; the proxy swaps it for a JWT-bearer employee token on `/v1/employee/*`.

## Notable

- The employee demo derives the bank substep from the flow bag's `selfOnboardingSubsteps` (not a direct hook call), so it needs **no extra SDK export** — consistent with dropping the `useGPOnboardingSteps` public export in #1142.
- Only the GP-relevant files are touched; stale demo-branch changes to unrelated example files were deliberately excluded.

## Verification

- Library build ✅ · example `type-check` (`tsc -b`) ✅ · oxfmt ✅ · oxlint ✅ (0 warnings on the demos)

---

> [!NOTE]
> **Medium Risk**
> Optional `legalEntityId` is a public API change for integrators; employment creation now depends on legal-entity fetch and new error states must be handled correctly.
> 
> **Overview**
> Adds a **GP Admin Onboarding** example demo that walks through country/basic info, contract, administrative details, and invitation, including handling for legal-entity load failures and missing GP-enabled entities.
> 
> **SDK:** `PayrollAdminOnboardingFlow` now wires in the real step components instead of placeholder null components. **`legalEntityId` is optional** — the hook always loads GP-enabled legal entities for the company, defaults to the first match when omitted, and surfaces `legalEntities` and `isErrorLegalEntities` on `adminBag` so hosts can distinguish fetch errors from “no GP legal entity.” Employment creation throws if no legal entity is available when creating a new employment.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46a9f85db979ea05d15331659b5551d5c4336cc2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/example/src/PayrollAdminOnboarding.tsx
+++ b/example/src/PayrollAdminOnboarding.tsx
@@ -1,0 +1,257 @@
+import { useState } from 'react';
+import {
+  PayrollAdminOnboardingFlow,
+  PayrollAdminOnboardingRenderProps,
+  useGPLegalEntities,
+} from '@remoteoss/remote-flows';
+import { RemoteFlows } from './RemoteFlows';
+import { AlertError } from './AlertError';
+import './css/main.css';
+
+const COMPANY_ID = import.meta.env.VITE_COMPANY_ID as string;
+
+const STEP_LABELS: Record<string, string> = {
+  select_country: 'Country & Basic Info',
+  contract_details: 'Contract Details',
+  administrative_details: 'Administrative Details',
+  invite: 'Send Invitation',
+};
+
+const STEP_DESCRIPTIONS: Record<string, string> = {
+  select_country:
+    "Select the employee's country and fill in their basic information.",
+  contract_details:
+    'Define the employment contract terms for this Global Payroll employee.',
+  administrative_details:
+    'Provide administrative details required for payroll processing.',
+  invite:
+    'Send the invitation email so the employee can complete their self-onboarding.',
+};
+
+type Errors = {
+  apiError: string;
+  fieldErrors: {
+    field: string;
+    messages: string[];
+    userFriendlyLabel: string;
+  }[];
+};
+
+const emptyErrors: Errors = { apiError: '', fieldErrors: [] };
+
+function AdminFlowForm({ legalEntityId }: { legalEntityId: string }) {
+  const [errors, setErrors] = useState<Errors>(emptyErrors);
+  const [done, setDone] = useState(false);
+
+  const clearErrors = () => setErrors(emptyErrors);
+
+  const handleError = (error: Error, fieldErrors: Errors['fieldErrors']) => {
+    setErrors({ apiError: error.message, fieldErrors });
+  };
+
+  if (done) {
+    return (
+      <div className='card' style={{ marginBottom: 20 }}>
+        <h1 className='heading'>Invitation Sent</h1>
+        <p style={{ color: '#22356f', marginBottom: 16 }}>
+          The employee will receive an email to complete their self-onboarding.
+        </p>
+        <div className='buttons-container'>
+          <button
+            className='submit-button'
+            onClick={() => {
+              setDone(false);
+              clearErrors();
+            }}
+          >
+            Start a new onboarding
+          </button>
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <PayrollAdminOnboardingFlow
+      companyId={COMPANY_ID}
+      legalEntityId={legalEntityId}
+      render={({ adminBag, components }: PayrollAdminOnboardingRenderProps) => {
+        const {
+          SelectCountryStep,
+          ContractDetailsStep,
+          AdministrativeDetailsStep,
+          InvitationStep,
+          SubmitButton,
+          BackButton,
+        } = components;
+
+        const currentStep = adminBag.stepState.currentStep.name;
+        const allSteps = Object.entries(STEP_LABELS);
+
+        if (adminBag.isLoading && !adminBag.countryCode) {
+          return <p>Loading...</p>;
+        }
+
+        const onStepError = (e: {
+          error: Error;
+          fieldErrors: { field: string; messages: string[] }[];
+        }) =>
+          handleError(
+            e.error,
+            e.fieldErrors.map((fe) => ({
+              ...fe,
+              userFriendlyLabel: fe.field,
+            })),
+          );
+
+        const renderStep = () => {
+          switch (currentStep) {
+            case 'select_country': {
+              const isFormReady =
+                !!adminBag.countryCode && adminBag.fields.length > 0;
+              return (
+                <>
+                  <SelectCountryStep
+                    onError={onStepError}
+                    onSuccess={clearErrors}
+                  />
+                  <AlertError errors={errors} />
+                  {isFormReady && (
+                    <div className='buttons-container'>
+                      <SubmitButton
+                        className='submit-button'
+                        onClick={clearErrors}
+                      >
+                        Create Employment & Continue
+                      </SubmitButton>
+                    </div>
+                  )}
+                </>
+              );
+            }
+            case 'contract_details':
+              return (
+                <>
+                  <ContractDetailsStep
+                    onError={onStepError}
+                    onSuccess={clearErrors}
+                  />
+                  <AlertError errors={errors} />
+                  <div className='buttons-container'>
+                    <BackButton className='back-button' onClick={clearErrors}>
+                      Previous Step
+                    </BackButton>
+                    <SubmitButton
+                      className='submit-button'
+                      onClick={clearErrors}
+                    >
+                      Save & Continue
+                    </SubmitButton>
+                  </div>
+                </>
+              );
+            case 'administrative_details':
+              return (
+                <>
+                  <AdministrativeDetailsStep
+                    onError={onStepError}
+                    onSuccess={clearErrors}
+                  />
+                  <AlertError errors={errors} />
+                  <div className='buttons-container'>
+                    <BackButton className='back-button' onClick={clearErrors}>
+                      Previous Step
+                    </BackButton>
+                    <SubmitButton
+                      className='submit-button'
+                      onClick={clearErrors}
+                    >
+                      Save & Continue
+                    </SubmitButton>
+                  </div>
+                </>
+              );
+            case 'invite':
+              return (
+                <>
+                  <AlertError errors={errors} />
+                  <div className='buttons-container'>
+                    <BackButton className='back-button' onClick={clearErrors}>
+                      Previous Step
+                    </BackButton>
+                    <InvitationStep
+                      onSuccess={() => {
+                        clearErrors();
+                        setDone(true);
+                      }}
+                      onError={onStepError}
+                    >
+                      Send Invitation
+                    </InvitationStep>
+                  </div>
+                </>
+              );
+            default:
+              return null;
+          }
+        };
+
+        return (
+          <>
+            <div className='steps-navigation'>
+              <ul>
+                {allSteps.map(([key, label], index) => (
+                  <li
+                    key={key}
+                    className={`step-item ${key === currentStep ? 'active' : ''}`}
+                  >
+                    <span className='step-number'>{index + 1}</span>
+                    {label}
+                  </li>
+                ))}
+              </ul>
+            </div>
+
+            <div className='card' style={{ marginBottom: 20 }}>
+              <h1 className='heading'>{STEP_LABELS[currentStep]}</h1>
+              <p style={{ fontSize: 14, color: '#71717A', marginBottom: 24 }}>
+                {STEP_DESCRIPTIONS[currentStep]}
+              </p>
+              {renderStep()}
+            </div>
+          </>
+        );
+      }}
+    />
+  );
+}
+
+function GPAdminOnboardingInner() {
+  const { data: legalEntities, isLoading } = useGPLegalEntities(COMPANY_ID);
+
+  if (isLoading) {
+    return <p>Loading…</p>;
+  }
+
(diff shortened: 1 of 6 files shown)


example/src/PayrollEmployeeOnboarding.tsx
```diff
@@ -0,0 +1,532 @@
...
+          (s: { type: string }) => s.type === 'employee_provides_bank_details',
+        );
+        const visibleSteps = Object.entries(STEP_LABELS).filter(([key]) => {
+          if (key === 'bank_account') return hasBankAccount;
+          if (key === 'federal_taxes') return isUSA;
+          if (key === 'state_taxes') return isUSA && !!jurisdiction;
+          return true;
+        });
+        const lastStepKey = visibleSteps[visibleSteps.length - 1][0];
+        const isLastStep = currentStep === lastStepKey;
```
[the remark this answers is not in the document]
hamzaremote: Fixed in 50f9594f. `lastStepKey` is now derived from the SDK-*reachable* steps (tax steps only count once `employeeBag.isComplete` is true — mirroring the SDK's own `buildEmployeeSteps` visibility), instead of the sidebar list. So on a pre-enrollment USA run the last step is correctly `bank_account` (or `home_address`), and Submit/`setDone` fire there.

example/src/PayrollEmployeeOnboarding.tsx
```diff
@@ -0,0 +1,532 @@
...
+                      >
+                        {isLastStep ? 'Submit' : 'Save & Continue'}
+                      </SubmitButton>
+                    ) : (
+                      <button
+                        className='submit-button'
+                        onClick={() => employeeBag.next()}
+                      >
+                        Skip
+                      </button>
```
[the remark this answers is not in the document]
hamzaremote: Fixed in 50f9594f. The federal_taxes 'Skip' now finishes (`setDone(true)`, labelled 'Finish') when it's the last reachable step — e.g. USA without a jurisdiction where there's no state-taxes step after it — matching the state-taxes path. It still just `next()`s when a later step exists.

example/src/PayrollEmployeeOnboarding.tsx
```diff
@@ -0,0 +1,548 @@
...
+        // fire on the real final step (e.g. bank_account on a pre-enrollment
+        // USA run).
+        const reachableSteps = Object.entries(STEP_LABELS).filter(([key]) => {
+          if (key === 'bank_account') return hasBankAccount;
+          if (key === 'federal_taxes') return isUSA && employeeBag.isComplete;
+          if (key === 'state_taxes')
+            return isUSA && !!jurisdiction && employeeBag.isComplete;
+          return true;
+        });
+        const lastStepKey = reachableSteps[reachableSteps.length - 1][0];
```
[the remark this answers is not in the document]
hamzaremote: Fixed in 0124a979. Completion is no longer latched from a stale `isLastStep` in the onSuccess closure. Each successful submit now records the step name (`setLastSubmittedStep(currentStep)`), and completion is derived per-render as `done || lastSubmittedStep === lastStepKey` — with `lastStepKey` recomputed each render from the SDK-reachable steps. So if `isComplete` flips after a bank save and tax steps become reachable, `lastStepKey` moves to the tax step, `lastSubmittedStep` (bank) no longer equals it, and the flow advances instead of completing. The explicit "Finish" on an unavailable tax step still sets `done`.

example/src/PayrollAdminOnboarding.tsx
```diff
@@ -0,0 +1,257 @@
...
+          AdministrativeDetailsStep,
+          InvitationStep,
+          SubmitButton,
+          BackButton,
+        } = components;
+
+        const currentStep = adminBag.stepState.currentStep.name;
+        const allSteps = Object.entries(STEP_LABELS);
+
+        if (adminBag.isLoading && !adminBag.countryCode) {
```
gabrielseco: isn't this a bit counterintuitive?

if countryCode is defined and isLoading is triggered by any step the isLoading, it won't show

example/src/PayrollAdminOnboarding.tsx
```diff
@@ -0,0 +1,257 @@
...
+            e.fieldErrors.map((fe) => ({
+              ...fe,
+              userFriendlyLabel: fe.field,
+            })),
+          );
+
+        const renderStep = () => {
+          switch (currentStep) {
+            case 'select_country': {
+              const isFormReady =
```
gabrielseco, line 135: shouldn't be isLoading and instead of not rendering the button put a disabled state?

example/src/PayrollAdminOnboarding.tsx
```diff
@@ -0,0 +1,257 @@
...
+            default:
+              return null;
+          }
+        };
+
+        return (
+          <>
+            <div className='steps-navigation'>
+              <ul>
+                {allSteps.map(([key, label], index) => (
```
gabrielseco, line 227: if for some reason in the future, the steps are dynamic like in the Onboarding & ContractorOnboarding, this won't work and we'll need to refactor it

Basically what we do is set the steps on the hook you read them through the adminBag and do the same iterations

example/src/PayrollAdminOnboarding.tsx
```diff
@@ -0,0 +1,257 @@
...
+            </div>
+          </>
+        );
+      }}
+    />
+  );
+}
+
+function GPAdminOnboardingInner() {
+  const { data: legalEntities, isLoading } = useGPLegalEntities(COMPANY_ID);
```
gabrielseco: I would solve this in the hooks internally.

If there are no legalEntities, this property can be exposed in the adminBag and they can do whatever they want with it

And the legalEntityId pass to the hook can be set internally, instead of expecting people setting them themselves.

if there are no legal entities nothing should be set

Let me know if I am missing anything

example/src/PayrollEmployeeOnboarding.tsx
```diff
@@ -0,0 +1,564 @@
+import { useState } from 'react';
```
gabrielseco, line 1: Can you extract this file to a separated PR

I want to review it standalone to avoid having mixed comments from the other files?

example/src/PayrollAdminOnboarding.tsx
hamzaremote: Fixed in 5d0fb9c5 + dafa5b95. Changed the gate to `adminBag.isLoading && adminBag.fields.length === 0` — it now blocks on any step whose schema hasn't loaded yet (not just before a country is picked), so it won't silently skip the loading state on step transitions later in the flow (e.g. while the contract_details/administrative_details schema is fetching, or while legal entities are being resolved).

example/src/PayrollAdminOnboarding.tsx
hamzaremote: Fixed in dafa5b95. The button is no longer conditionally unmounted — it's always rendered with `disabled={!isFormReady}` (SubmitButton also composes that with its own isSubmitting/isLoading disabling), so it stays visible and just goes disabled/enabled as readiness changes.

example/src/PayrollAdminOnboarding.tsx
hamzaremote: Good callout — leaving this as static `STEP_LABELS`/`STEP_DESCRIPTIONS` for now since the admin flow's steps aren't dynamic today (select_country is the only conditionally-visible one, and this demo never sets skipCountry). Agreed on the approach if/when that changes: read the steps through adminBag the same way Onboarding/ContractorOnboarding do, rather than hardcoding them in the demo. Not doing that refactor in this PR since it's speculative for the current flow.

example/src/PayrollAdminOnboarding.tsx
hamzaremote: Implemented in dafa5b95 — moved this into `usePayrollAdminOnboarding` itself: `legalEntityId` is now optional, and when not provided the hook fetches the company's GP-enabled legal entities via `useGPLegalEntities` internally and defaults to the first one. `adminBag.legalEntities` exposes the (possibly empty) array so the consumer can render its own "no legal entity" state — the demo now does exactly that instead of a separate wrapper component that fetched legal entities up front. If `legalEntityId` still ends up unset when submitting select_country, `onSubmit` throws rather than silently sending an empty ID.

example/src/PayrollEmployeeOnboarding.tsx
hamzaremote: Done — split out to #1201, which now carries PayrollEmployeeOnboarding.tsx plus its supporting server-side auth plumbing (jwt_auth.js employee-assertion minting, proxy.js routing, the employee-token route) and the two follow-up fix commits that were already reviewed here. This PR is now scoped to just the admin demo.

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -45,12 +48,21 @@ const buildAdminSteps = (
...
   options,
 }: Omit<PayrollAdminOnboardingFlowProps, 'render'>) => {
+  // Only fetch when the caller hasn't already pinned a legal entity — avoids
+  // an unnecessary request for callers that already know theirs.
+  const { data: legalEntities, isLoading: isLoadingLegalEntities } =
+    useGPLegalEntities(providedLegalEntityId ? undefined : companyId);
+
+  // If there are no GP-enabled legal entities, this stays undefined — the
+  // consumer is expected to check `legalEntities` and not render the flow.
+  const legalEntityId = providedLegalEntityId ?? legalEntities?.[0]?.id;
```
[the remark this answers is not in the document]
hamzaremote: Fixed in 815ec667 — removed the fetch-skip optimization. usePayrollAdminOnboarding now always calls useGPLegalEntities(companyId) regardless of whether legalEntityId was provided explicitly, so adminBag.legalEntities stays accurate for the "empty means no GP legal entity" contract either way. Added a test covering the explicit-legalEntityId + populated-legalEntities case.

jordividaller (approved):

src/flows/PayrollAdminOnboarding/hooks.tsx
```diff
@@ -45,12 +48,23 @@ const buildAdminSteps = (
...
+  // Always fetch, even when the caller already pinned a legal entity — so
+  // `legalEntities` below stays accurate for callers that inspect it (e.g. to
+  // render their own "no GP-enabled legal entity" state) instead of always
+  // reporting empty just because the fetch was skipped.
+  const { data: legalEntities, isLoading: isLoadingLegalEntities } =
+    useGPLegalEntities(companyId);
+
+  // If there are no GP-enabled legal entities, this stays undefined — the
+  // consumer is expected to check `legalEntities` and not render the flow.
+  const legalEntityId = providedLegalEntityId ?? legalEntities?.[0]?.id;
```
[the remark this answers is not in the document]
hamzaremote: Fixed in a9635dd5 — `usePayrollAdminOnboarding` now also exposes `isErrorLegalEntities` (from `useGPLegalEntities`'s `isError`), and the demo checks it before falling back to the empty-legalEntities message, rendering a distinct error state instead. Updated the `legalEntityId` JSDoc to point consumers at it, and added a hook test covering the fetch-failure case.

example/src/PayrollEmployeeOnboarding.tsx
```diff
@@ -0,0 +1,564 @@
...
+                All your information has been submitted. Your employer will
+                review and activate your employment.
+              </p>
+              <div className='buttons-container'>
+                <button className='submit-button' onClick={startOver}>
+                  Start over
+                </button>
+              </div>
+            </div>
+          );
```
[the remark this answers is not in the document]
hamzaremote: This file is no longer part of this PR — it was split out to #1201, which already merged into `main`. I also found and fixed a related rebase issue on this branch: after rebasing onto `main` (which now includes #1201), a leftover commit here was re-deleting the already-merged employee-onboarding demo and its auth plumbing; that's been corrected so this PR is scoped to admin-onboarding only and doesn't touch `PayrollEmployeeOnboarding.tsx` at all.

The underlying "Start over skips flow reset" bug you're flagging is real but now lives in `main` via #1201, not in this PR's diff — worth tracking as a separate follow-up against the employee demo rather than blocking this PR.

gabrielseco (approved):

PR #1184, opened 2026-07-23
fix(employment-agreements) - fix the step when devs skip country

gabrielseco:
We didn't check that the feature worked when devs use `skipSteps=['select_country'] `

Fix that and also fix examples to allow reuse, some tests testing the behavior and reorder a bit the stucture of the folder to allow reusage a bit


jordividaller (approved):
Thanks for reordering example folders it is really better like this :).

PR #1201, opened 2026-07-28
feat(gp): example app demo for employee self-onboarding flow (PBYR-4044)

hamzaremote:
## Summary
- Split out of #1171 per review request, so this demo can be reviewed standalone without comments on the admin demo mixed in.
- Adds the Global Payroll employee self-onboarding demo (`example/src/PayrollEmployeeOnboarding.tsx`) and its server-side auth plumbing (employee-assertion JWT minting in `jwt_auth.js`, path-based token routing in `proxy.js`, the `/api/fetch-employee-token/:employmentId` route).
- Includes the two follow-up fixes already reviewed/applied on #1171: correct last-reachable-step detection and reactive (non-stale-closure) completion tracking.

## Test plan
- [x] `npm run type-check`
- [x] `npm run lint`
- [x] `npm run check-format`
- [x] `npx vitest run` (823 tests passing)
- [ ] Manual click-through of the demo in a browser (reviewer to verify, needs sandbox credentials)

---

> [!NOTE]
> **Medium Risk**
> Introduces server-side OAuth JWT minting and path-based token selection in the example API; mitigated by production 403 guards and confinement to the example dev proxy, but it still handles client secrets and broad employee write scopes.
> 
> **Overview**
> Adds a **GP Employee Self-onboarding** example demo wired to `PayrollEmployeeOnboardingFlow`, including step navigation, USA tax availability handling, and completion logic that tracks the **last reachable** step (so tax steps appearing after enrollment do not prematurely finish the flow).
> 
> **Dev-proxy auth** is extended so `/v1/employee/*` calls use an **employment-scoped JWT-bearer** minted server-side: the proxy reads `x-rf-employment-id`, exchanges a signed assertion for an access token, and strips that header before forwarding. A non-production **`/api/fetch-employee-token/:employmentId`** route mirrors the existing company-manager helper. **`RemoteFlows`** gains **`authType='none'`** so the browser does not fetch tokens when the proxy handles auth per path.
> 
> The demo uses a **two-layer** setup: an outer context loads employment country/jurisdiction via `/v1/employments/*` (user-token path through the proxy), then an inner context sends employee API traffic with the employment header only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68a4bb3a931f3acf96f0e837aa9f2a7bcd531fe8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/example/api/proxy.js
+++ b/example/api/proxy.js
@@ -3,25 +3,30 @@ const {
   fetchClientCredentialsAccessToken,
   fetchAccessToken,
 } = require('./get_token.js');
+const { fetchEmployeeToken } = require('./jwt_auth.js');
 const { buildGatewayURL } = require('./utils.js');
 
 /**
  * Determines which token type to use based on HTTP method and path
  * @param {string} method - HTTP method (GET, POST, PUT, PATCH, etc.)
  * @param {string} path - The API path (e.g., '/v1/countries' or '/v2/countries?foo=bar')
- * @returns {'client-credentials' | 'user-token'} The token type to use
+ * @returns {'client-credentials' | 'user-token' | 'employee-assertion'} The token type to use
  */
 function getTokenType(method, path) {
   const normalizedMethod = method.toUpperCase();
   // Extract pathname without query parameters
   const pathname = path.split('?')[0].toLowerCase();
 
-  // GET /v1/countries or /v2/countries
+  // GET /v1/countries or /v2/countries — these don't require a user identity;
+  // use client_credentials so the call works in CI where no user token is
+  // available. Local dev without a client secret can opt into a user token
+  // by setting VITE_REMOTE_GATEWAY=... and ensuring VITE_CLIENT_TOKEN works.
   if (normalizedMethod === 'GET' && /^\/v[12]\/countries$/.test(pathname)) {
     return 'client-credentials';
   }
 
-  // GET /v1/countries/{country_code}/address_details or /v2/countries/{country_code}/address_details
+  // GET /v[12]/countries/{country_code}/address_details — public reference
+  // data; also use client credentials.
   if (
     normalizedMethod === 'GET' &&
     /^\/v[12]\/countries\/[^/]+\/address_details$/.test(pathname)
@@ -48,6 +53,13 @@ function getTokenType(method, path) {
     /^\/v[12]\/companies\/[^/]+$/.test(pathname)
   ) {
     return 'client-credentials';
+  }
+
+  // /v1/employee/* endpoints need an employment-scoped assertion. The FE
+  // identifies which employment via the x-rf-employment-id header; the proxy
+  // mints the JWT-bearer token server-side so the FE never sees it.
+  if (/^\/v1\/employee\//.test(pathname)) {
+    return 'employee-assertion';
   }
 
   // All other requests use user token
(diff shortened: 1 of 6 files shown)


example/api/proxy.js
```diff
@@ -50,6 +55,13 @@ function getTokenType(method, path) {
     return 'client-credentials';
   }
 
+  // /v1/employee/* endpoints need an employment-scoped assertion. The FE
+  // identifies which employment via the x-rf-employment-id header; the proxy
+  // mints the JWT-bearer token server-side so the FE never sees it.
+  if (/^\/v1\/employee\//.test(pathname)) {
+    return 'employee-assertion';
+  }
+
```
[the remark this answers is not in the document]
hamzaremote: Fixed in 68a4bb3a — corrected the comments/error text rather than actually wiring company-manager routing in proxy.js. Reason: GET /v1/employments/{id} is also called by ContractorOnboarding, Termination, ContractAmendment, and Onboarding through this same shared dev proxy, none of which are part of this PR — changing getTokenType's routing there would change their auth behavior too, which needs its own review/verification against the gateway rather than a silent side-effect here. The comments/error message now accurately describe what actually happens today: this outer context resolves auth via the same refresh-token ('user-token') flow as most other demos behind the proxy, not a dedicated company-manager mint (authType on the FE component is moot here since the proxy middleware always overwrites the outgoing Authorization header itself).

jordividaller (approved):

example/src/PayrollEmployeeOnboarding.tsx
```diff
@@ -0,0 +1,571 @@
...
+    </div>
+  );
+}
+
+export function PayrollEmployeeOnboardingForm() {
+  return (
+    // Outer context: authType='none' since the FE doesn't hold a token here
+    // either — see the useEmployeeFlowContext comment above for how auth is
+    // actually resolved for /v1/employments/* through the dev proxy.
+    <RemoteFlows proxy={{ url: window.location.origin }} authType='none'>
```
gabrielseco, line 567: All the logic has to converge in one provider and not in two

First we need to understand authType='none' wasn't used until now, we used fetchCompanyToken which is the same thing that we do with authType='none' but users don't see the accessToken returned.

We have a form where we ask about an employmentId and then in the example we use useEmploymentQuery -> this should be inside the PayrollEmployeeOnboardingFlow

Then you can derive from the employmentId the countryCode and jurisdiction

Once you have that you can pass inside the PayrollEmployeeOnboardingFlow to the employee calls the header 

`headers: { 'x-rf-employment-id': employmentId },`

Partners won't like this complexity that we're doing and we need to hide it

hamzaremote: understood, I will follow up with an MR

hamzaremote: Confirmed — the proxy routes purely by path (`/v1/employee/*` → employee-assertion, everything else → user-token), so the dual-provider split wasn't actually needed. Moved the fix into the SDK flow itself rather than the example: `PayrollEmployeeOnboardingFlow`/`usePayrollEmployeeOnboarding` now derive `countryCode`/`jurisdiction` internally via `useEmploymentQuery(employmentId)` and attach `x-rf-employment-id` directly on the employee-scoped mutations. The example now uses a single `RemoteFlows` provider.

Opened as #1204 (targets `main`, since this needed to land in the SDK flow, not just this PR's example-app diff).

